PR #11, opened 2015-10-13
Django 1.6 compatibility for views

mbetz08:
More efforts to achieve Django 1.6 compatibility!

## Problem ##
Graphene's Django views [uses Django JsonResponse object](https://github.com/graphql-python/graphene/blob/master/graphene/contrib/django/views.py#L3), which wasn't [introduced until Django 1.7.](https://docs.djangoproject.com/en/1.8/ref/request-response/#jsonresponse-objects)

## Proposed Solution ##
Use HttpResponse instead of JsonResponse with JSON encoded data.  Also, tests that passed on master passed with this change.

### DIFF
--- a/graphene/contrib/django/views.py
+++ b/graphene/contrib/django/views.py
@@ -1,6 +1,6 @@
 import json
 
-from django.http import JsonResponse
+from django.http import HttpResponse
 from django.views.generic import View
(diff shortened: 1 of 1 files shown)


syrusakbary:
:+1:

PR #12, opened 2015-10-15
Test resolver tags w/ tag_resolver and resolver_has_tag function.

jhgg:
And some other clean-ups because I'm slightly OCD.


syrusakbary:
Thanks!

Issue #18, opened 2015-11-04
Improvements next stable package


syrusakbary:
Inspired by other GraphQL implementations, next version should be able to:

* [x] Simplify Fields definition (rename `graphene.{type}Field` to `graphene.{type}`).
* [x] Add Types dynamic instantiation into ObjectTypes, Interfaces (as field), or in Field kwargs (as Arguments)
* [x] Add `.List` , `.NonNull` to Types
* [x] Add Union Type
* [x] Add Scalar Type (for having, for example a DateTime Scalar type)

jhgg:
Mmmmmmmm. Good ideas.

PR #21, opened 2015-11-12
Ride to 0.4.0

syrusakbary:
Based in #18, this are the things that will be improved:

Inspired by other GraphQL implementations, next version should be able to:

* Simplify Fields definition (rename `graphene.{type}Field` to `graphene.{type}`).
* Add Types dynamic instantiation into ObjectTypes, Interfaces (as field), or in Field kwargs (as Arguments)
* Add `.List` , `.NonNull` to Types
* Add Union Type
* Add Scalar Type (for having, for example a DateTime Scalar type)

Also, any non-specific fields are now marked as deprecated, such as `StringField`, `IDField`, `IntField`, `FloatField`, `BooleanField`, `ListField` and `NonNullField`.


Globegitter:
Looking forward to seeing this land :)

PR #25, opened 2015-11-13
Use `graphql-django-view` to handle `GraphQLView`

### LINKED ISSUE #23: django GraphQLView isn't implemented properly
https://github.com/graphql-python/graphene/blob/master/graphene/contrib/django/views.py

Isn't implemented properly. Doesn't do proper error handling & doesn't handle GraphQL variables at all.

I was thinking instead we could remove this from graphene, and make this into it's own standalone package that could be given a GraphQL Schema and create a Django View for it. 

That way it could be used with bare `graphql-core`, `graphql-epoxy` and `graphene`.

Thoughts?

jhgg:
Fixes #23.

### DIFF
--- a/graphene/contrib/django/views.py
+++ b/graphene/contrib/django/views.py
@@ -2,9 +2,15 @@ from graphql_django_view import GraphQLView as BaseGraphQLView
 
 
 class GraphQLView(BaseGraphQLView):
+    graphene_schema = None
+
     def __init__(self, schema, **kwargs):
         super(GraphQLView, self).__init__(
+            graphene_schema=schema,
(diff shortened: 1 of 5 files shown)


syrusakbary:
@jhgg If you can change the branch to merge into to `0.4.0` instead of `master` would be great!

syrusakbary:
And... deleting repeated tests maybe?

jhgg:
Yeah. I will clean out tests. Just wanted to leave them there for now to prove that it would be a plug-in replacement.

jhgg:
I'll have to open a new PR to change branch to merge.

PR #36, opened 2015-11-18
Add the info parameter (ResolveInfo) to get_node() calls.

### LINKED ISSUE #34: Need to have resolution "helpers" (including in get_node)
In my preferred SQLAlchemy setup, simply having the model class is not sufficient to load a model. In addition, you must have a separate `session` object, which has a query method.

For example, the Django approach of :
```python
ship = Ship.objects.get(id=1)
```

would instead work like this:
```python
ship = session.query(Ship).filter_by(id=1).one()
```

I also require the ability to pass a notion of the current user into the query so that my resolvers can verify the user is authorized to access this data.

This means the 'session' and 'auth' objects must be provided to the `schema.execute` method and get passed into any resolvers. I had planned to do this using the ExecutionContext. However, the `get_node` classmethods don't get provided the ExecutionContext.

What do you suggest?

inklesspen:
This allows request_context (or any other ResolveInfo data) to be used while getting nodes.
For example, some data might need to be hidden based on the user's authorization; you would use info.request_context for this.

Fixes #34.

### DIFF
--- a/examples/starwars_django/schema.py
+++ b/examples/starwars_django/schema.py
@@ -17,7 +17,7 @@ class Ship(DjangoNode):
         model = ShipModel
 
     @classmethod
-    def get_node(cls, id):
+    def get_node(cls, id, info):
         return Ship(get_ship(id))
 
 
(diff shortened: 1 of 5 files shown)


syrusakbary:
Great! Merging.... will add a function inspector on top for assure compatibility (checking if the function has one or two args, and decorating with a warning for the old-style `get_node`.

PR #39, opened 2015-11-20
Example of querying a Python object

amitsaha:
While getting started, I did not find a simple example of what I wanted to do here - query a Python object type and not a scalar field. Hopefully this is a good addition.

### DIFF
--- a/examples/field_example.py
+++ b/examples/field_example.py
@@ -28,4 +28,4 @@ query = '''
 '''
 result = schema.execute(query)
 # Print the result
-print result.data['patron']
+print(result.data['patron'])
--- a/examples/field_example.py
+++ b/examples/field_example.py
@@ -1,0 +1,30 @@
+import graphene
+
+class Person(graphene.Interface):
+    name = graphene.String()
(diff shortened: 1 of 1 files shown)


syrusakbary:
I like the example @amitsaha, thanks for contributing!
The `graphene` project runs flake8 syntax checker, and there are some syntax improvements that needed to be made in order to pass flake8 (and pass the tests).
https://travis-ci.org/graphql-python/graphene/jobs/92172692#L337

amitsaha:
Thanks for taking a look @syrusakbary , fixed the issues now.

amitsaha:
@syrusakbary related question, is *this* the simplest way to do this? Do I have always have to define an interface and then subclass it ? Can't I use a simple class instead as a field?

syrusakbary:
@amitsaha nope, you can directly subclass from `graphene.ObjectType` ;)

amitsaha:
OK, that's great. I have changed the example now. Please take a look thanks.

syrusakbary:
Good work!

Issue #48, opened 2015-11-23
Django integration proposal


adamcharnock:
I've been thinking a bit about Django integration for Graphene. This is in part because I think it could be really useful, and in part because this is something I'd be interested in contributing a pull request for. However, development seems to be moving quickly so I thought it best to discuss these ideas first in case there are already plans.

The main motivation here is reducing boilerplate code. My intention would be to make all automatic functionality overridable & customisable.

**Nodes:**

-  ~~Automatically traverse Django relations. A `resolve_FIELD()` *may* be specified to customise this traversal, but if not then traversal will be done using the relation's manager's `all()` method.~~ [already supported]
- [x] Automatically support query args for all/specified model fields (reduces `resolve_FIELD()` boilerplate code).

**Query:**

- [x] Automatically resolve `ConnectionFields` which connect to `DjangoNodes` without the need for a `resolve_FIELD()` method.
- [ ] Support for pulling in schema definitions from individual Django apps, rather than being defined centrally. (I currently do this by using a top-level `Query` class which inherits from multiple app-specific `Query` classes) [related: #55]

**Mutations:**

- [ ] Provide a `DjangoMutation` which provides standard functionality though which to update Django models.
- [ ] Consider a class-based views approach, with `CreateMutation`, `UpdateMutation`, and `DeleteMutation` base classes.

All/any thoughts very welcome indeed!

jhgg:
One cool thing would be to inspect field asts and generate the correct prefetch_related, select_related and only calls onto the queryset. I think syrus is working on something like this.

adamcharnock:
I agree @jhgg, I think that is #28.

As far as my ability to contribute is concerned, I would probably look to implement the basic ideas/functionality which could then be optimised.

syrusakbary:
Right now graphene is already transversing Django relations (only if the target model is in some `ObjectType` in the schema), and resolving to ConnectionFields whenever the target model is defined in some `DjangoNode` in the schema, without the need for a `resolve_FIELD()` method.

I was thinking about adding support to [django-filter](https://github.com/alex/django-filter) for filtering the querysets (so you can define which attributes you can filter on).
I agree that could be very useful having the possibility to autodiscover the schema by searching into the django `INSTALLED_APPS` (something similar to django.admin is doing).

About optimizing the hits to the dbs only fetching the data needed... I'm working on that! :)

The mutations will probably require a little bit more of work, but will be very useful having those!

syrusakbary:
[...]

PR #60, opened 2015-12-02
Django integration implementation

adamcharnock:
I'm creating this (work in progress) pull request given that #48 seems to have developed a sensible solution.

This pull request currently represents the contents of the original proposed Gist which can be found at:
https://gist.github.com/adamcharnock/ad051b419d4c613d40fe

Todo:

- [x] Create a `GlobalIDFilter` and `GlobalIDField` field
-  ~~Consider allowing the new resolvers to be used as function decorators (as per @syrusakbary suggestion)~~
- [x] Allow FilterSet classes to be created dynamically created upon passing a `fields` argument to `DjangoFilterConnectionField` (as per @syrusakbary suggestion)
- [x] Write tests
- [x] `django-filter` should be an optional package
- [x] Filtering on to-many fields
- [x] Change order parameter name to `order` rather than `o`
- [x] Filtering information should be stored on the destination node's `Meta` class
- [x] Filtering should be available on nodes without having to explicitly define the fields
- ~~Filtering should be enabled explicitly (i.e. via Django setting) rather than implicit (via presence of `django-filter`).~~
- [x] Documentation

Useful, but perhaps for a second phase of development:

- [ ] Support for lazy loading of related nodes
- [ ] Loading of app-level query classes


syrusakbary:
I've updated graphene with a big refactor using `ClassTypes` (shouldn't suppose a problem for fixing the merge issues, but ping me if any question!)

Apart from that I think `django-filter` should not be a required package by default, instead I think we should only allow the usage of the filter fields if `django-filter` is present (throwing an error otherwise).

Other than this and your checklist items everything looks good!
(I will have to spend more time seeing the tests and the usage for having more insights!)

adamcharnock:
Great, I've just rebased the branch. There was only a minor conflict in `graphene/contrib/django/__init__.py`.

I was wondering the same thing about `django-filter`, I've added it to the list.

I've got some client work coming up next week, so my *hope* is to have this mostly done in the next few days.

adamcharnock:
@syrusakbary I'm thinking about your idea of method decorators. I'm imaging something like this:

```python
class Query(ObjectType):
    all_accounts = DjangoFilterConnectionField(AccountNode)

    @filter_connection(fields=['id', 'name', 'slug'], order_by=['name', 'slug'])
    def resolve_all_accounts(self, args, info):
        return Account.objects.filter(user=info.context.request_context.user)
```

Is that roughly what you had in mind?

Is there any benefit of doing that over doing this (which is now implemented):

```python
class Query(ObjectType):
    all_accounts = DjangoFilterConnectionField(AccountNode, 
                                               fields=['id', 'name', 'slug'], 
                                               order_by=['name', 'slug'])

    def resolve_all_accounts(self, args, info):
        return Account.objects.filter(user=info.context.request_context.user)
```

syrusakbary:
I was thinking about something like you proposed, however I just realized the Field should know which arguments it's going to receive (you cannot send random arguments, as have to be created explicitly), so the arguments have to be typed in the Field definition. Adding then `@filter_connection` decorator only adds redundancy.

So, let's keep it lean and simple, and go with your current implementation without method decorators :)

adamcharnock:
Any views on the ordering field being called `order` or `order_by`? It is the former for now. I felt django-filter's default of `o` was too terse for the otherwise more descriptive language used in GraphQL.

syrusakbary:
I've been thinking more, and maybe the filter args could be inside a InputObjectType, so instead of requesting:
```graphql
{
   posts(first:2, headline__icontains: "peter") {
      title
   }
}
```
We can do 
```graphql
{
   posts(first:2, filter: {headline__icontains: "peter"}) {
      title
   }
}
```

This way, we can implement exclude very easily too. However I'm not very convinced yet.
Thoughs?

syrusakbary:
I think having a `order_by` arg is fine! :+1:

syrusakbary:
How mount works:

All the fields when, contributed to a class (meaning, when placed as attrs inside a class), hold a relation to the class where is contributed.
Like:
```python
class MyObjectType(ObjectType):
    name = Field(String())
```
So the `Field(String())` is now referencing it's parent `MyObjectType` (because it was contributed to it).
However, the inner types (in our case `String()`) have to hold a reference to `MyObjectType`, so in this case (for inner types), we use `.mount()`.

**But, Why referencing or mounting the inner fields?**
It's important to hold this reference somewhere, because in some cases we need to access it, like:

```python
class Human(ObjectType):
    friends = Field(List('self'))
```

Hope this clarifies a little bit how mounting works! @adamcharnock

adamcharnock:
Thanks for that @syrusakbary, that helps a lot!

I think your idea of putting the filter args inside a InputObjectType makes technical sense. However, I really like how cleanly the GraphQL reads in the current system. I haven't actually made much use of this filtering in my Relay app yet, so I'm not sure what the impact will be from a front-end code point of view. I'm tempted to 'wait and see' on this one. What do you think?

adamcharnock:
I'm also having [a lot of trouble](https://github.com/adamcharnock/graphene/commit/64ec0ca94fc576bf56af508ce90cb3330b502c9d#diff-4cda9bab71a0819b2c8096ff3740c4f3R161) trying to test [some new code](https://github.com/adamcharnock/graphene/commit/64ec0ca94fc576bf56af508ce90cb3330b502c9d#diff-445735212f9936ce318fb3f988d2e020R60) ([see in coverage.io](https://coveralls.io/builds/4348492/source?filename=graphene%2Fcontrib%2Fdjango%2Ffilter%2Ffilterset.py#L60)). It seems that Django hasn't loaded the relations for the models and therefore I cannot navigate reverse relations in the tests. This probably isn't surprising given we are testing outside of Django's testing framework, but it is causing me a problem.

Do you have any suggestions for a way around this?

syrusakbary:
@adamcharnock for testing with Django db initialized (and setup'ed) you can write this mark from `pytest-django`, in the tests:

```python
import pytest

pytestmark = pytest.mark.django_db
```

This is how [starwars_django](https://github.com/graphql-python/graphene/blob/master/examples/starwars_django/tests/test_objectidentification.py) is dealing with it. Let me know if that works ;)

adamcharnock:
I've updated the Django version requirement to >=1.8. It seems that the tests on master don't pass for 1.7 or 1.6, and fixing the issues proved difficult. I've also updated the travis config to build both 1.8 and 1.9. How do people feel about this?

adamcharnock:
@syrusakbary Also, thank you for the tip on the test! That helped a lot :-)

graphene/contrib/django/form_converter.py
```diff
@@ -0,0 +1,70 @@
...
+
+@convert_form_field.register(forms.DecimalField)
+@convert_form_field.register(forms.FloatField)
+def convert_form_field_to_float(field):
+    return Float(description=field.help_text)
+
+
+@convert_form_field.register(forms.ModelMultipleChoiceField)
+@convert_form_field.register(GlobalIDMultipleChoiceField)
+def convert_form_field_to_list_or_connection(field):
```
syrusakbary: `convert_form_field_to_list` instead?

graphene/contrib/django/form_converter.py
```diff
@@ -0,0 +1,70 @@
...
+    return List(ID())
+
+
+@convert_form_field.register(forms.ModelChoiceField)
+@convert_form_field.register(GlobalIDFormField)
+def convert_form_field_to_djangomodel(field):
```
syrusakbary: `convert_form_field_to_id`

graphene/contrib/django/form_converter.py
adamcharnock: Good point, done

graphene/contrib/django/form_converter.py
adamcharnock: Good point, done

syrusakbary:
I know some `graphene` developers that are using Django 1.6, so I would like to still support them.
I can work with you in fixing those tests (even if we add some boilerplate code) ;)

You know why the tests seem to fail? Which build? Maybe not testing in Python 3.3+Django 1.9 fixes the tests?

adamcharnock:
It seems that there are differences in the Django ORM API in different versions, certainly involving relations. My head is a bit fuzzy at this point though. Definitely up for graphene working in order django versions.

adamcharnock:
I've also just added this in 3a23c1f940fbd30293b5722c872daa9a86c4c924:

```python
schema = graphene.Schema(name='Cookbook Schema')


class CategoryNode(DjangoNode):
    class Meta:
        model = Category
        filter_fields = ['name', 'ingredients'] # Can now specify filtering on the destination node


class IngredientNode(DjangoNode):
    class Meta:
        model = Ingredient
        filter_fields = ['name', 'notes', 'category']


class Query(ObjectType):
    category = relay.NodeField(CategoryNode)
    all_categories = DjangoFilterConnectionField(CategoryNode)

    ingredient = relay.NodeField(IngredientNode)
    all_ingredients = DjangoFilterConnectionField(IngredientNode)

schema.query = Query
```

Having this option available makes sense to me as the attributes a node can be filtered on are primarily related to the node, not the connection. They can still be specified/overridden at the connection level though.

adamcharnock:
I'm thinking that creating a `DjangoFilterNode` class could make the most sense. This way I can customise how the fields get generated* and also create a separate `DjangoFilterOptions` class to handle the additional `Meta` options that will be needed.

Very much open to thoughts on any of this though.

* as per the PR's todo item: 'Filtering should be available on nodes without having to explicitly define the fields'

syrusakbary:
I really like your suggestions.

I think we should remain as lean as possible, that means if we can include the `filter_fields` attr in the classes `graphene.contrib.django.types.ObjectType` and `graphene.contrib.django.types.Node` would be better (and simpler) than create `Filter...` classes.

So instead of using `DjangoFilterNode`, use `DjangoNode`.
But only permit to set the `filter_fields` attr if is supported by the environment (meaning the `django-filter` package is installed).

Implementing this logic in the django `ObjectTypeMeta` will work better, as `DjangoNodeMeta` inherits from it, and also the `fitler_fields` attrs could be set in django `ObjectType`.

Apart from that, I think we should create also a `DjangoFilterField` class, that acts as a Field with a `List`, but with the filter args (as `DjangoFilterConnectionField`).

Thoughts @adamcharnock ? :)

syrusakbary:
However, for keeping the logic separated, you can create a separate meta class and let django `ObjectTypeMeta` extend it (but only if the `django-fitler` is found).

graphene/contrib/django/filter/filterset.py
```diff
@@ -0,0 +1,115 @@
...
+
+class GlobalIDMultipleChoiceFilter(MultipleChoiceFilter):
+    field_class = GlobalIDMultipleChoiceField
+
+    def filter(self, qs, value):
+        gids = [from_global_id(v).id for v in value]
+        return super(GlobalIDMultipleChoiceFilter, self).filter(qs, gids)
+
+
+ORDER_BY_FIELD = getattr(settings, 'GRAPHENE_ORDER_BY_FIELD', 'order')
```
syrusakbary: I'm more inclined to use `order_by` as I automatically think of it as a String (field to order), as opposed as using `order` that I think more as a Boolean (if we want to order the results: `True` or `False`).

adamcharnock: Done, bbbf6884492360e6ae1b57263d02f067b18c7d7c

adamcharnock:
@syrusakbary Done, with caveats:

1. I implemented your last point regarding `DjangoObjectTypeMeta`. However, I actually created a `DjangoFilterObjectTypeMeta` class which is an optional base of `DjangoNodeMeta`. I needed this in order to adjust the default django field conversion. I'm not sure this is a good idea, partly because it looks weird, and party because...
2. ... [two tests are breaking](https://travis-ci.org/graphql-python/graphene/jobs/95045987#L578) and I'm not sure why. If `DjangoNodeMeta` does not inherit from `DjangoFilterObjectTypeMeta` then they pass.

Thoughts very welcome!

graphene/contrib/django/types.py
```diff
@@ -46,6 +50,15 @@ def construct(cls, *args, **kwargs):
         return cls
 
 
+class DjangoFilterObjectTypeMeta(ObjectTypeMeta):
+
+    def convert_django_field(cls, field):
+        from graphene.contrib.django.filter import DjangoFilterConnectionField
+        field = super(DjangoFilterObjectTypeMeta, cls).convert_django_field(field)
+        field.connection_field_class = DjangoFilterConnectionField
+        return field
```
adamcharnock: Thoughts on this new `connection_field_class` and its use?

syrusakbary: @adamcharnock Why we have to convert the django field?

adamcharnock:
Ok, I'll take a look.

Any thoughts on those broken tests?

adamcharnock:
@syrusakbary Ok, I've reverted the commit. I think my concern was more to do with behaviour during imports. But I think you are right.

syrusakbary:
I will take a look tomorrow on the broken tests.

adamcharnock:
Thank you :+1: I may be less around this week depending on how busy I am with work.

adamcharnock:
So by now it is probably clear that I haven't had time to revisit this yet! It sounds like I'll be working full-time in January too.

However, I'd really like to get this merged in so it can start being useful. Aside from documentation, is there anything that needs to be done before it can be merged? ( @syrusakbary )

syrusakbary:
Nope, I can improve things also after merge @adamcharnock. However documentation could be really useful

adamcharnock:
@syrusakbary Great. I've done a little more work on the docs but then got sidetracked by a bug I discovered. It seems that the results are not correctly returned when relationships are traversed (for filtered models only). All results are returned rather than just the results for that relationship.

I've spent a couple of hours trying to figure this out, but I've clearly forgotten how a lot of this hangs together since I last came to it. Would you be able to shed any light on this?

I'll do some more work on the docs.

syrusakbary:
I will try to fix the issue with the relations, the test you added really helps! ;)
If you can focus on the documentation would be great! :)

adamcharnock:
@syrusakbary Great, thank you!

I've just expanded on the django quckstart and added a separate section on filtering. Let me know if you think I've missed anything (or it needs any amends), but other than that I think I'm happy to consider it done.

adamcharnock:
I've also added an example django app named `cookbook`. I left the `starwars` app intact for a couple of reasons:

1. I felt it could still have value in its current form
2. On a philosophical note: I know SW is the canonical GraphQL example, but I feel that understanding it requires some specific pop culture knowledge. Also SW fans are predominantly white males. Therefore, in the interest of diversity and common knowledge I went with food :-)

syrusakbary:
I have to re-read everything and add some test cases. But this could be merged very soon @adamcharnock :)

adamcharnock:
Brilliant! Thanks so much for taking a look over it.

michaelkuty:
:+1:

syrusakbary:
I've tested this among with the changes in #79 and this is now good to go.
@adamcharnock @michaelkuty

adamcharnock:
Fantastic! :dancer:

PR #64, opened 2015-12-06
Add plugin to debug Django SQL queries in the Query (similar to DjangoDebugToolbar)

syrusakbary:
@adamcharnock could you take a look into that?

I'm working in a new plugin system for Graphene that permits to access or modify some internal very easily.
I would like to have some feedback before releasing the code to the public.

Thanks!!


graphene/contrib/django/debug/plugin.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def debug_objecttype(objecttype):
+    return type(
+        'Debug{}'.format(objecttype._meta.type_name),
+        (WrapRoot, objecttype),
+        {'debug': Field(DjangoDebug, name='__debug')})
+
+
+class DjangoDebugPlugin(Plugin):
```
syrusakbary: Here is where the Plugin inits

graphene/contrib/django/debug/tests/test_query.py
```diff
@@ -0,0 +1,70 @@
...
+          }
+          allReporters {
+            lastName
+          }
+          __debug {
+            sql {
+              rawSql
+            }
+          }
+        }
```
syrusakbary: ^ Here is how to do the GraphQL query for see all the sql queries executed

adamcharnock:
Hey @syrusakbary, thanks for pinging me. I think allowing modification of internal functionality is a great idea.

The way I've seen this done in the past is:

1. Configuring which classes/modules should be used to provide functionality (such as Django's `DEFAULT_FILE_STORAGE`, `TEMPLATE_LOADERS` etc). I think this can break down into:
   * Specifying a single class/module/etc which should be used (`DEFAULT_FILE_STORAGE`)
   * Specifying multiple classes/modules in more of a 'behaviour pattern' style (`TEMPLATE_LOADERS`, `INSTALLED_APPS`)
2. Providing hooks into internal code. I've normally seen this done explicitly in the code which is to be modified. (Django [has signals](https://docs.djangoproject.com/en/1.9/topics/signals/), wordpress has [apply_filters](http://codex.wordpress.org/Function_Reference/apply_filters) and [do_action](http://codex.wordpress.org/Function_Reference/do_action), [used like this](https://github.com/WordPress/WordPress/blob/77e365efbf2e499e2ed11d29c101ea466cf1ceed/wp-admin/user/admin.php#L27))
3. Monkey-patching (which I'd say is best avoided)

After a little reading, I think the characteristics of a good plugin system are:

1. Easy to understand and use (i.e. both in code, and in documentation)
2. Loosely coupled. There is some central plugin manager with code code on one side, and plugin code on the other. Each side does not know or care about the other
3. Easy to debug
4. Have a mechanism for registering plugins (and potentially inspecting those plugins installed)

**Easy to understand**

I've been hacking around with Graphene for about a week now and I'm still pretty unsure of how the schema & type system work. I don't think this is the fault of Graphene, but rather this is just a project that is both inherently complex and lacking in documentation of that complexity (for now).

If the plugin system needs to be easier to understand than the graphene internals, then **perhaps the plugin API should be more verbose**. For example, providing multiple clearly named hooks.

I feel the code in the `CamelCase` plugin requires I knowledge of Graphene's internals. One must know that is it `GroupNamedType.get_named_type()` that one needs to override, which I don't think would be easy to discover. Neither `GroupNamedType` or `get_named_type` shout out to me as obvious places for this to be.

Separately: The `CamelCase` plugin also needs to check the type of `_type` for each call. I feel this is something Django's signals [have a solution to](https://docs.djangoproject.com/en/1.9/topics/signals/#connecting-to-specific-signals). (I'm not saying we should use Django's signals, rather that this is an idea to consider)

**Loosely coupled**

*I say this based on [a blog post from 2008](http://martyalchin.com/2008/jan/10/simple-plugin-framework/), but I don't this has changed much.*

The current plugin seems to not obey this principle as plugins require knowledge of Graphene internals.

**Easy to debug**

Plugins start to change code behaviour in unexpected ways. Even if *how* the behaviour will change cannot be made clear, I think the fact that behaviour *can* change should be. In the current plugin system it seems that we monkey patch a method to customise the functionality. I think this is a problem because:

1. The original code will never be called and it would not be obvious why to a developer doing debugging
2. There are no hints in the core code that it's behaviour may be changed

I'd suggest something like:

```python
class GroupNamedType(BaseType):
    ...
    def get_named_type(self, schema, type):
        name = type.name or type.attname
        # Either:
        name = signals.type_name.trigger(name, sender=GroupNamedType)
        # Or:
        name = PluginManager.trigger("get_type_name", name)
        return name, schema.T(type)
```

... or something along those lines.

Node that Django's signals will return a list of values, one for each handler, so this would need to change in our implementation.

**Mechanism for registering plugins:**

This is currently done by passing `Schema(plugins=[...])`, which seems like a reasonable solution to me.

I'm think I'm out of thoughts for now :-) Thoughts?

syrusakbary:
Thanks for such a detailed response!

My thoughts... I always think as signals as something potentially `async`. I feel little bit weird doing a trigger of a signal for retrieving their response in a synchronous way. But the concept of how it should be done is good.

You are right about all the points you mentioned. A plugin should be easy to debug, and mutating the object_type in real time it's probably not the best thing to do.

I will improve the code soon based on your suggestions!


### Offtopic: Explanations on internal code

*Graphene internals have to be improved. Adding internal `__docs__`, and probably some refactor in the code for making it easier to understand probably are musts in the future.*

Right now I'm more focused in how I want to use graphene.

Even after 3 major refactors on the internals, the syntax have remained the same (which is great!)

> You can improve the internals whenever you want if the public facing API is well tested, but the syntax is what will make the usage for working on that easier.

A very interesting related video conference: [simplicity is complicated](http://www.thedotpost.com/2015/11/rob-pike-simplicity-is-complicated).

However, the more simplicity we can achieve also in the code, the better. And any suggestions for improving there will be always welcome :)

syrusakbary:
@adamcharnock the related PR to camelCase plugin: https://github.com/graphql-python/graphene/pull/65

Issue #46, opened 2015-11-23
Snake case to camel case


millerjs:
I didn't find a forced conversion to camelCase in the spec. Did I miss it? Or is graphene just opinionated on style standards.  Will there be any way to opt out of this in the future?

jhgg:
It's convention. Generally you would want to use camel case in javascript world. But graphene keeps things snake cased in Python. If you want to use snake cased fields though you can explicitly specify a name.

syrusakbary:
Hi @millerjs, thanks for asking!
I think path to go will be having *snake_cased* fields converted to *camelCase* by default, but adding the possibility for changing this via some global variable or graphene settings.
Will keep this issue open until graphene adds the ability of not camelCase if defined in settings.

syrusakbary:
Probably adding custom permissions for querying/updating using `django-guardian` is also a good enhancement (and could be required when using automutation models?). Thoughts about that?

millerjs:
:+1: This would be a great feature despite it's unconventionality.

syrusakbary:
@millerjs this is now fixed.
Thanks to a new plugin structure you can even create [some plugin for transform the name](https://github.com/graphql-python/graphene/blob/master/graphene/plugins/camel_case.py) the way you want.

The current usage for disable the auto camelCase is:
```python
schema = graphene.Schema(query=..., auto_camelcase=False)
```

Hope it helps! ;)

millerjs:
Fantastic! thanks @syrusakbary

sulabhjain:
> @millerjs this is now fixed.
> Thanks to a new plugin structure you can even create [some plugin for transform the name](https://github.com/graphql-python/graphene/blob/master/graphene/plugins/camel_case.py) the way you want.
> 
> The current usage for disable the auto camelCase is:
> 
> ```python
> schema = graphene.Schema(query=..., auto_camelcase=False)
> ```
> 
> Hope it helps! ;)

Hi @syrusakbary, I am using this thread for a question related to auto_camelcase, let me know if there is more appropriate forum to addresses this issue.

We are running into a requirement where not all clients of graphql api wants to use camelCase for interacting with the api (eg. non-javascript based internal services). So the question here is, if there is a way to dynamically pass a parameter to graphql api which determines if camelCase to snake_case conversion should happen or not on the server side. If not, is this a requirement on the future roadmap ? Appreciate your help.

oxbits:
defaulting to camel case is not a fair assumption

[...]

PR #65, opened 2015-12-07
Create plugin structure

syrusakbary:
This PR creates a basic plugin structure and adds the ability to not `camelCase` by default the field/argument names as exposed in #46


adamcharnock:
[[Initial conversation](https://github.com/graphql-python/graphene/pull/64#issuecomment-162347013), #64]

No problem, it's nice to look into a new problem. I've added a couple of comments into the commits, but overall I think this is great. I think figuring out where the hooks should be could be a larger job.

I don't think I considered signals to be async, but I presume that is just because were coming from different backgrounds. I think your view is valid.

I'm also slightly aware that I don't seem to have seen this kid of hook system used in many other python projects. I'm wondering if this is just my experience or if there is some larger reason. I'd be interested to get other opinions on this.

Re 'Explanations on internal code': Sounds perfectly reasonable.

syrusakbary:
@adamcharnock I've improved everything based on your considerations :)

PR #67, opened 2015-12-10
Optimize Django SQL queries based on the GraphQL queries fields

syrusakbary:
As GraphQL provides a way for fetching only certain fields, the Django mapper should adapt for querying only what is requested. That means using `model.objects.only(*requested_fields)` as will be much more performant.


graphene/contrib/django/fetcher.py
```diff
@@ -0,0 +1,28 @@
...
+        _type = _type.of_type
+
+    for field in field_asts:
+        field_def = get_field_def(info.schema, _type, field)
+        f = field_def.resolver
+        fetch_field = getattr(f, 'django_fetch_field', None)
+        if not fetch_field:
+            continue
+        only_args.append(fetch_field.attname)
+    return only_args
```
bigblind: This function seems useful outside Django, to get the fields from an info object. Is there any way this could be refactored into  graphene core? I'm willing to do the work, but I'm not familiar enough with the codebase to know where this should fit.

Globegitter: Good catch @bigblind - that would be great to have in graphene core/`ResolveInfo`

mjtamlyn:
My understanding is that `.only()` is in fact a performance hit in many cases. This may seem counter intuitive, but it comes down the the fact that the returned model object from an only query is much more complex than a raw one, and this extra time is a bigger hit that the extra work your database does.

See [the documentation](https://docs.djangoproject.com/en/1.9/ref/models/querysets/#defer) for only and defer, the note pretty much says "don't do this unless you absolutely have to".

ProjectCheshire:
@phalt this is when django was under graphene.contrib.django :) I the faq / upgrading it was moved in v1.0 to a seperate package

phalt:
@ProjectCheshire guess we can close it then!

PR #69, opened 2015-12-11
Add documentation about using custom input object types as arguments

ustun:
This document is based on my reading of the tests.

Ideally, without reading the test, I attempted to create a custom type using normal means, and providing it as an argument again as normal means.

After lots of head scratching, I found the solution, but maybe the behaviour could be fixed at the library level so that the following works:
```python
class CustomArgumentObjectType(graphene.ObjectType):
    field1 = graphene.String()
    field2 = graphene.String()
```
And then:

```
graphene.Field(graphene.String(), to=CustomArgumentObjectType())

```


syrusakbary:
Thanks for the PR @ustun, I think it would help a lot to people who's starting with Graphene!
I will think more about having `InputObjectTypes` as `Arguments` when used in `Fields` without specifically writing it!

PR #73, opened 2015-12-18
Implement Django command for dumping schema.json

lightning18:
* Implement Django `./manage.py graphql_schema`

* Required to have 'graphene.django.contrib' in INSTALLED_APPS setting

* Introduces two new optional settings GRAPHENE_SCHEMA and
  GRAPHENE_SCHEMA_OUTPUT

* GRAPHENE_SCHEMA for the Django app containing the schema

* GRAPHENE_SCHEMA_OUTPUT for the output file (defaults to schema.json)

* Use `./manage.py graphql_schema -h` for help in command line

* Also a handy solution for issue #3

### DIFF
--- a/graphene/contrib/django/management/commands/graphql_schema.py
+++ b/graphene/contrib/django/management/commands/graphql_schema.py
@@ -1,0 +1,38 @@
+from django.core.management.base import BaseCommand, CommandError
+
+import importlib
+import json
+
+
+class Command(BaseCommand):
+    help = 'Dump Graphene schema JSON to file'
+    can_import_settings = True
+
+    def add_arguments(self, parser):
+        from django.conf import settings
+        parser.add_argument(
+            '--schema',
+            type=str,
+            dest='schema',
+            default=getattr(settings, 'GRAPHENE_SCHEMA', ''),
+            help='Django app containing schema to dump, e.g. myproject.core.schema')
+
+        parser.add_argument(
+            '--out',
+            type=str,
+            dest='out',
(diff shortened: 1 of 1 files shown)


syrusakbary:
Great work! :+1:

adamcharnock:
@lightning18 This is great, thank you :-)

Just a minor hiccup: It seems that the `SUCCESS` style was [only added in Django 1.9](https://github.com/django/django/commit/f06ce6053cae6a96ecce29ce2020789945757bf4). I don't really see a sensible alternative style though :s

lightning18:
That is unfortunate, is it a deal breaker? Should I use an if statement to support 1.8 too?

adamcharnock:
I recall @syrusakbary stating a preference to support Django 1.6 as quite a few of the graphql-core folks are using it. IMHO I don't think it is a critical problem in this case, but it is a very useful command so it would be great to sort it somehow.

Personally, I'd be tempted to just remove the styling.

Side note – I didn't know that Django offered this styling - I'm sure I'll be making use of it soon :-)

lightning18:
Yes, sure, I don't mind it being remove, it's just a visual after all. I can submit a PR for that. I'd also like to include some documentation. Should I just add it on the Django quickstart?

syrusakbary:
I'm working in fixing the compatibility in older versions of Django. And adding some basic testing for the command.

If you could add some documentation would be great! @lightning18 (just create a dedicated .md file in [this folder](https://github.com/graphql-python/graphene/tree/master/docs/pages/docs) and I will take care of the rest)

Issue #76, opened 2016-01-07
Incorrect mutate document


youngking:
[This page] (http://graphene-python.org/docs/mutations/)  has a mutate example,  but parameters of the  mutate method are incorrect, follow it will get an error:  `"mutate() takes exactly 3 arguments (4 given)" `

According to  to this [test](https://github.com/graphql-python/graphene/blob/master/graphene/core/tests/test_mutations.py), it should be like:

```python
    @classmethod
    def mutate(cls, instance, args, info):
        person = Person(name=args.get('name'))
        ok = True
        return CreatePerson(person=person, ok=ok)
```

syrusakbary:
This is solved by #77.

PR #77, opened 2016-01-07
Changed mutation documentation in reference to Issue #76

BillyBarbaro:
I believe the function listed in the tutorial is incorrect as mentioned in issue #76.  Changed it to match what's used in the Star Wars examples.

### DIFF
--- a/docs/pages/docs/mutations.md
+++ b/docs/pages/docs/mutations.md
@@ -22,7 +22,7 @@ class CreatePerson(graphene.Mutation):
     person = graphene.Field('Person')
 
     @classmethod
-    def mutate(cls, args, info):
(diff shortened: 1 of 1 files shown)


syrusakbary:
Hi @BillyBarbaro, the problem is in the function arguments, not in the function name (`mutate_and_get_payload` is only for the Relay mutations).
Instead of `def mutate(cls, args, info):` should be `def mutate(cls, instance, args, info):`

BillyBarbaro:
@syrusakbary Thanks for the correction there.  I've only ever used graphene alongside Relay, but that all makes sense now.  Made the correction.  Let me know if you think there's anything else that needs to be done.

syrusakbary:
:+1:

PR #81, opened 2016-01-15
Fixed NodeField object_type string name argument.

glowka:
Hey, I found bug related to ```relay.NodeField('MyNode')``` behavior, which I assume should be exactly the same as using ```relay.NodeField(MyNode)```. Here is simple fix for that.


syrusakbary:
Great work! Thanks for the patch ;)

PR #99, opened 2016-02-01
Fix UnionType resolver

### LINKED ISSUE #98: Example for UnionType
We have just been trying to use `UnionType` but with very little success, all throwing some error message that is hard to identify/debug.

This is a simplified example of what we have been trying, using the latest version `0.7.0`:

```py
from graphene.core.classtypes.uniontype import UnionType
from graphene import ObjectType, String, Field, Schema

class Human(ObjectType):
    some_field = String()
    other_field = String()

class Pet(ObjectType):
    yet_another_field = String()

class Thing(UnionType):
    '''Thing union description'''
    class Meta:
        types = [Human, Pet]

class Query(ObjectType):
    thing = Field(Thing)

    def resolve_thing(self, args, info):
        return Thing(some_field='test') # or Human(some_field='test')

```

There are a few more variations that we have tried as well, but didn't work. So how does a full example of using `UnionType` look like?

CC @imranolas

Globegitter:
Fixes #98

### DIFF
--- a/graphene/core/classtypes/uniontype.py
+++ b/graphene/core/classtypes/uniontype.py
@@ -35,6 +35,6 @@ class UnionType(six.with_metaclass(UnionTypeMeta, FieldsClassType)):
         return GraphQLUnionType(
             cls._meta.type_name,
             types=list(map(schema.T, cls._meta.types)),
-            resolve_type=cls._resolve_type,
+            resolve_type=lambda instance, info: cls._resolve_type(schema, instance, info),
             description=cls._meta.description,
         )
--- a/graphene/core/classtypes/uniontype.py
+++ b/graphene/core/classtypes/uniontype.py
@@ -1,3 +1,5 @@
+from functools import partial
+
 import six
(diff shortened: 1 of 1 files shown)


graphene/core/classtypes/uniontype.py
```diff
@@ -35,6 +35,6 @@ def internal_type(cls, schema):
         return GraphQLUnionType(
             cls._meta.type_name,
             types=list(map(schema.T, cls._meta.types)),
-            resolve_type=cls._resolve_type,
+            resolve_type=lambda instance, info: cls._resolve_type(schema, instance, info),
```
syrusakbary: For being in constance with the [implementation of `Interface`](https://github.com/graphql-python/graphene/blob/master/graphene/core/classtypes/interface.py#L51), maybe the following will be better.

```python
from functools import partial
...
            resolve_type=partial(cls._resolve_type, schema),
```

syrusakbary:
Good catch!

Globegitter:
Should be update now :)

Issue #101, opened 2016-02-02
Calculated fields


sgtsquiggs:
How do we access calculated fields? In a serializer I would do something like
```python
class ArtworkSerializer(serializers.ModelSerializer):
    url = serializers.URLField(source='get_absolute_url')

    class Meta:
        model = Artwork
        fields = ('id', 'title', 'url')
```

I assume I need to specify this somehow in `class ArtworkNode(DjangoNode):` but I've been unable to discover how.

sgtsquiggs:
Figured it out.

```python
class ArtworkNode(DjangoNode):
    url = graphene.String()

    class Meta:
        model = Artwork
        filter_fields = ['id', 'title']

    def resolve_url(self, args, info):
        return self.instance.url
```


Is it possible to retrieve via `pk`?

syrusakbary:
Great! However I'm going to add the source option to fields! It makes it much more pleasant to develop!

So you can do something like

```python
class ArtworkNode(DjangoNode):
    url = graphene.String(source='get_absolute_url')

    class Meta:
        model = Artwork
        filter_fields = ['id', 'title']
```

syrusakbary:
@sgtsquiggs You can retrieve an Artwork Node doing:
```
{
  node(id:"SOME_STRING") {
    id
  }
}
```

In this case `SOME_STRING` will be the value of `ArtworkNode.global_id(1)`.

sgtsquiggs:
Is it possible to get a single item via a field other than id? for instance I have a `slug` field on Artwork and in my django endpoint I use Artwork.get_by_slug(slug).

`source=` would be perfect, I use it extensively in my serializers

syrusakbary:
Sure. You can do is using arguments.

```python
class Query(graphene.ObjectType):
    artwork = graphene.Field(ArtworkNode, slug=graphene.String())
    
    def resolve_artwork(self, args, info):
        return Artwork.get_by_slug(args.get('slug', None))
```

Then you can query it with the following query:
```
{
  artwork(slug:"my_slug") {
    id
    title
  }
}
```
Hope it helps!

paymog:
How do I resolve a django object along with an extra, calculated field like below during a query?

```python
class ProjectType(DjangoObjectType):
    class Meta:
        model = Project

    extra_field = graphene.String()


class Query(ObjectType):
    project = graphene.Field(ProjectType,
                             name=graphene.String())

    def resolve_project(self, info, name):

        if id is not None:
[...]

PR #102, opened 2016-02-03
Added source argument in Fields

syrusakbary:
As reflected in issue #101, this pull request adds the ability on `Field`s to support mapping to an `instance` attribute different than the field name, allowing the following to be possible:

```python
class ArtworkNode(DjangoNode):
    url = graphene.String(source='get_absolute_url')

    class Meta:
        model = Artwork
        filter_fields = ['id', 'title']
```

### DIFF
--- a/graphene/core/classtypes/objecttype.py
+++ b/graphene/core/classtypes/objecttype.py
@@ -46,6 +46,9 @@ class ObjectType(six.with_metaclass(ObjectTypeMeta, FieldsClassType)):
     class Meta:
         abstract = True
 
+    def __getattr__(self, name):
+        return self._root and getattr(self._root, name)
+
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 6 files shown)


Speedy1991:
Can you give an example how this works?
e.g.: This does not work:

        ls = graphene.String(source='_ls', path=graphene.String())

        @staticmethod
        def _ls(path=None):
            cmd = "ls -la "
            return subprocess.check_output((cmd + path) if path else cmd, shell=True).decode("utf-8")

Running `{ls}` leads to:
```
{
  "data": {
    "ls": null
  }
}
```

With a default resolver it just runs fine

jkimbo:
@Speedy1991 `source` is specific to the [graphene-django](https://github.com/graphql-python/graphene-django) integration.

For plain graphene you can use the `resolver` kwarg to specify a custom resolver: http://docs.graphene-python.org/en/latest/types/objecttypes/#resolvers-outside-the-class

PR #104, opened 2016-02-03
Add a doc on authorization in Django

defrex:
This distills what I had to figure out to limit data access in Graphene. Since it's such a common task and isn't very intuitive, I figured it'd be a good idea to add some documentation.


syrusakbary:
That's a great contribution!! Merging :+1:

Issue #105, opened 2016-02-03
Error trying to create a custom Scalar type


defrex:
I'm trying to create a field type called `Money` that is a decimal in python, and rounds to cents as an integer when it's serialized. As per the docs, this is my class.

```python
from decimal import Decimal

from graphene.core.classtypes import Scalar
from graphql.core.language import ast


class Money(Scalar):
    @staticmethod
    def serialize(value):
        return int(value * Decimal(100))

    @staticmethod
    def parse_literal(node):
        if isinstance(node, ast.IntValue):
            return Decimal(node) / Decimal(100)

    @staticmethod
    def parse_value(value):
        return Decimal(value) / Decimal(100)
```

Given this `ObjectType`.

```python
class Summary(graphene.ObjectType):
    income = graphene.Field(Money())
    allocated = graphene.Field(Money())
    spent = graphene.Field(Money())
    net = graphene.Field(Money())
```

I get the following error.

```
Traceback (most recent call last):
  File "/home/defrex/code/spendwell/apps/users/tests.py", line 16, in test_safe_to_spend
    result = self.graph_query('{ viewer { safeToSpend } }', user=owner)
  File "/home/defrex/code/spendwell/apps/core/tests.py", line 21, in graph_query
    result = schema.execute(query, request_context=request)
  File "/home/defrex/code/spendwell/.env/lib/python3.5/site-packages/graphene/core/schema.py", line 123, in execute
    kwargs = dict(kwargs, request=request, root=root, args=args, schema=self.schema)
  File "/home/defrex/code/spendwell/.env/lib/python3.5/site-packages/graphene/core/schema.py", line 86, in schema
    subscription=self.T(self.subscription))
  File "/home/defrex/code/spendwell/.env/lib/python3.5/site-packages/graphene/core/schema.py", line 22, in __init__
    super(GraphQLSchema, self).__init__(*args, **kwargs)
  File "/home/defrex/code/spendwell/.env/lib/python3.5/site-packages/graphql/core/type/schema.py", line 42, in __init__
    self._type_map = self._build_type_map()
  File "/home/defrex/code/spendwell/.env/lib/python3.5/site-packages/graphql/core/type/schema.py", line 91, in _build_type_map
    type_map = type_map_reducer(type_map, type)
  File "/home/defrex/code/spendwell/.env/lib/python3.5/site-packages/graphql/core/type/schema.py", line 131, in type_map_reducer
[...]

PR #107, opened 2016-02-04
Add some clarifications to the custom Scalar type doc

defrex:
This tripped me up when trying to create a custom type. I also had to hunt in the source for the `Scalar` import so I added that as well.

### DIFF
--- a/docs/pages/docs/basic-types.md
+++ b/docs/pages/docs/basic-types.md
@@ -39,6 +39,7 @@ If you want to create a DateTime Scalar Type just type:
 
 ```python
 import datetime
+from graphene.core.classtypes import Scalar
 from graphql.core.language import ast
 
 class DateTime(Scalar):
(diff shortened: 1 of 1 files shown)


syrusakbary:
Hi @defrex, thanks for the PR!

I'm actually thinking of allowing DateTime() types inside Fields.
I will either merge your PR or merge something that allows doing what you described in #105

defrex:
Consistency with the other basic types seems like a good idea, though if `graphene.Field` is still necessary that should probably be noted in the docs.

If you let me know where you net out I can adjust the PR accordingly.

syrusakbary:
Great. I just fixed, so you can also do instances of Scalars as Fields or Arguments, like this:

```
class Post(graphene.ObjectType):
    created = DateTime()
```

defrex:
Awesome. In that case, maybe it's not necessary to show off the custom field being used. It's still a good idea to add the import though.

syrusakbary:
Great!
I will merge once the tests are complete.

PR #109, opened 2016-02-05
Fix incorrect type in example

timmygee:
The ok field in the example should be a Boolean since it is treated as one


syrusakbary:
That's right! Merging!

PR #127, opened 2016-03-06
Allow empty list in SQLAlchemyConnectionField.

miracle2k:
Custom resolve()'s empty list return value should take precedence over SQLAlchemyConnectionField default query.

I have the following:
     
    class Client(relay.Node):
        menuitems = SQLAlchemyConnectionField(MenuItem, connection_type=ProductsOnClient)
         def resolve_menuitems(self, args, info):
            return self.menu_items
 
Without this patch, if ``self.menu_items`` is an empty list, it will be ignored and I get items from other clients.

### DIFF
--- a/graphene/contrib/sqlalchemy/fields.py
+++ b/graphene/contrib/sqlalchemy/fields.py
@@ -17,7 +17,7 @@ class SQLAlchemyConnectionField(ConnectionField):
         return self.type._meta.model
 
     def get_query(self, resolved_query, args, info):
-        return resolved_query or get_query(self.model, info)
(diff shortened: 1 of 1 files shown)


syrusakbary:
Good catch!
This was fixed in Django with the commit https://github.com/graphql-python/graphene/commit/314703d7b5380c80fe2ee1e2ad4a3eb3d9ed7fc4#diff-233528cf66f4e9df736a367fe12b2ec1R30 but forgot to replicate in SQLAlchemy.

syrusakbary:
I will probably improve it later using the default method, as keep things simpler!

Issue #145, opened 2016-04-12
Use other name than id as primary key in SQLAlchemy models


karec:
Hi,

Simple question:  is it possible to use an other attribute than ID as primary key with SQLAlchemy plugin ?

It's seems to not be implemented yet : 

```python
class SQLAlchemyNode(six.with_metaclass(
        SQLAlchemyNodeMeta, NodeInstance)):

    class Meta:
        abstract = True

    @classmethod
    def get_node(cls, id, info=None):
        try:
            model = cls._meta.model
            query = get_query(model, info)
            instance = query.filter(model.id == id).one()
            return cls(instance)
        except NoResultFound:
            return None
```

If I understand this code correctly it seems that a node will always filter on the `id` attribute of the model. It would be realy nice and useful to have a meta information to set the identifier of a model for using other columns as primary key

syrusakbary:
Currently we use the id, but I'm open for define a more flexible way for approaching this!

karec:
Working on it right now ;)

karec:
#146

karec:
I close it since the PR has been merged :)

PR #146, opened 2016-04-13
Add meta attribute identifier for SQLAlchemyNode

karec:
Add ability to specify custom primary key column in `SQLAlchemyNode` through an `identifier` attribute in the meta class of it


karec:
Reference to issue #145

syrusakbary:
Looks great! 👍 

Merging (in few hours a new `graphene` version will be in pypi)

PR #152, opened 2016-04-29
improve iso8601 handling

jdugan1024:
The DateTime scalar accepts only a small portion of valid ISO8601 date time strings.  This pull request uses the iso8601 library to improve that.

As an example here's a Date created in node.js and rendered as an ISO8601 string:

```js
> var d = new Date()
Fri Apr 29 2016 13:34:12 GMT-0500 (CDT)
> d.toISOString()
'2016-04-29T18:34:12.502Z'
```

This fails with the old parsing method:

```python 
import datetime
iso_from_node = '2016-04-29T18:34:12.502Z'
# this line is copied from the old parsing code
datetime.datetime.strptime(iso_from_node, "%Y-%m-%dT%H:%M:%S.%f")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/_strptime.py", line 328, in _strptime
    data_string[found.end():])
ValueError: unconverted data remains: Z

import iso8601
iso8601.parse_date(iso_from_node)
datetime.datetime(2016, 4, 29, 18, 34, 12, 502000, tzinfo=<iso8601.Utc>)
```

I have included tests as well.  I wasn't able to run the full test suite because following the instructions in the README.md file didn't work in a fresh virtualenv didn't work.  I suspect it can be made to work but I don't have the time right this moment.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,6 +57,7 @@ setup(
         'six>=1.10.0',
         'graphql-core>=0.4.9',
         'graphql-relay==0.3.3',
+        'iso8601==0.1.11',
     ],
     tests_require=[
         'django-filter>=0.10.0',
--- a/setup.py
+++ b/setup.py
@@ -57,7 +57,7 @@ setup(
         'six>=1.10.0',
         'graphql-core>=0.4.9',
         'graphql-relay==0.3.3',
-        'iso8601==0.1.11',
+        'iso8601',
     ],
     tests_require=[
         'django-filter>=0.10.0',
--- a/graphene/core/types/custom_scalars.py
+++ b/graphene/core/types/custom_scalars.py
@@ -1,5 +1,6 @@
 import datetime
 import json
+import iso8601
 
 from graphql.core.language import ast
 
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -57,6 +57,7 @@ def run_tests(self):
         'six>=1.10.0',
         'graphql-core>=0.4.9',
         'graphql-relay==0.3.3',
+        'iso8601==0.1.11',
```
syrusakbary: I will either have the `iso8601` to be in `graphql/core/pyutils/iso8601.py` and import from there, or to require a generic version of `iso8601` in setup.py (generic version == no version specified).

The reason for that is any other package could be already using an older or newer version of `iso8601` and when installing `graphene` package it will trigger a incompatibility error as requires a specific version of it. (That's the reason Django includes a lot of third-party libraries in their own codebase).

jdugan1024: @syrusakbary OK.  I'm fine removing the version limitation there -- it's just habit for me to put it in there. I can see how in this case where you have a library used by lots of different projects that you don't want to lock down the version number.  I'll update the setup.py file.

PR #155, opened 2016-05-06
Updated references/imports for latest version graphql-core

AlecAivazis:
@syrusakbary: as I mentioned in slack, this PR updates the references/imports of `graphql-core` to accommodate the new api. 

Unfortunately, I'm having difficulty running the test suite (I think it's related to the version downloaded when the tests run). I will try to figure out the problem later today when I get some downtime.

### DIFF
--- a/graphene/core/schema.py
+++ b/graphene/core/schema.py
@@ -66,8 +66,6 @@ class Schema(object):
 
     @property
     def executor(self):
-        if not self._executor:
-            self._executor = SyncExecutor()
         return self._executor
 
     @executor.setter
--- a/graphene/core/schema.py
+++ b/graphene/core/schema.py
@@ -1,12 +1,10 @@
 import inspect
 from collections import OrderedDict
 
-from graphql.core.execution.executor import Executor
-from graphql.core.execution.middlewares.sync import \
-    SynchronousExecutionMiddleware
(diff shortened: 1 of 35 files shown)


graphene/core/schema.py
```diff
@@ -69,8 +67,7 @@ def T(self, _type):
     @property
     def executor(self):
         if not self._executor:
-            self._executor = Executor(
-                [SynchronousExecutionMiddleware()], map_type=OrderedDict)
+            self._executor = SyncExecutor()
```
AlecAivazis: This is the only part that seems to have drastically changed - I did some investigating and it looks like the Executor uses an OrderedDict by default now so I don't think I needed to specify it explicitly.

syrusakbary: We were using `self._executor` because the default used in graphql-core in versions previous to `0.5.0` was not using a OrderedDict by default.
However this is no longer necessary, as now we can use the default chosen by `graphql-core`.
Just doing `return self._executor` should work fine :)

AlecAivazis: Ah, got it. I removed that logic - still having trouble with the tests, though. I'll give it another shot later today

syrusakbary:
I've created the `core-update` branch in graphene based on your work with fix in tests and other enhacements.
Closing this PR, and opening from the `core-update` branch :)

PR #162, opened 2016-05-13
fix impossible dependencies of dependencies

### LINKED ISSUE #161: graphene[django] has clashing dependencies
This becomes evident when using pip-tools (pip-compile) to check dependencies.
Because of ``graphql-django-view==1.3b1`` it ends up trying to install
``graphql-core>=0.5.0==0.5b3`` which is impossible.

stefanfoulis:
fixes #161


syrusakbary:
Good catch. Thanks! :)

PR #166, opened 2016-05-17
Pass context to resolve methods

### LINKED ISSUE #164: Discussion - Pass context to resolve methods
The context story seems very confused in the newest version. My understanding from [the graphql-js 0.5.0 release notes](https://github.com/graphql/graphql-js/releases/tag/v0.5.0) is that all resolving functions should now have the signature `([self,] args, context, info)`, as opposed to `([self,] args, info)`.

At the moment, if I update a `resolve_NAME` method to take context, then it throws argument errors.

Also, mutations don't take context and therefore they cannot access context data. It think a patch similar to [this one](https://github.com/graphql/graphql-relay-js/pull/77) on graphql-relay-js is necessary for the relay mutations at least, but probably for all of them.

danielfarrell:
Also had to add a few deps to tox.ini to make tests run/pass

fixes #164

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,10 +36,10 @@ class Query(graphene.ObjectType):
     ping = graphene.String(description='Ping someone',
                            to=graphene.String())
 
-    def resolve_hello(self, args, info):
+    def resolve_hello(self, args, context, info):
         return 'World'
 
-    def resolve_ping(self, args, info):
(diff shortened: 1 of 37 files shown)


danielfarrell:
Should be handled using the `with_context` decorator.

PR #171, opened 2016-05-19
Fix invalid references to User model in Django tutorial

iwaszko:
I've found and fixed some issues in Django tutorial available at http://graphene-python.org/docs/django/tutorial/. I guess some comments were referring to an older code example.


syrusakbary:
Thanks!

PR #175, opened 2016-05-20
Allow numbers in to_constant

AdrielVelazquez:
Enums aren't effected if a int is passed along with the string.

### DIFF
--- a/graphene/utils/str_converters.py
+++ b/graphene/utils/str_converters.py
@@ -18,4 +18,4 @@ def to_snake_case(name):
 
 
 def to_const(string):
-    return re.sub('[\W|^(?=\d)]+', '_', string).upper()
+    return re.sub('[\W|^]+', '_', string).upper()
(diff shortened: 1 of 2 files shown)


syrusakbary:
Looks good for me, please fix the tests so I can merge:
https://travis-ci.org/graphql-python/graphene/jobs/131773091#L448

AdrielVelazquez:
@syrusakbary  All set!

syrusakbary:
@AdrielVelazquez 

Another thing, for assuring that everything work... could you add this test too?
(*all numbers starting in a string should be skipped*).

```python
def test_to_const_skip_first_char_if_number():
    assert to_const('13 snakes $1. on a "#plane') == 'SNAKES_1_ON_A_PLANE'
```

AdrielVelazquez:
@syrusakbary to_const is being used to convert Django choices to constant values. Django currently accepts any value for these Enums even if starting with integers. This also is the case with MySQL and Postgres. A starting integer wouldn't effect the current functionality of the dictionary being created (or rather the list created to guarantee unique type)

I can add additional tests of Django models with choices getting translated correctly and choices unaffected?

syrusakbary:
I was worried about Enum names, not values... but that should work.
Merging!

PR #178, opened 2016-05-22
Fixed bug when no middlewares are present

AlecAivazis:
I ran into an issue with schemas when `auto_camelcase` was set to `False` and no middlewares were set on the resolver. This resulted in an unnamed field which could not be executed.


syrusakbary:
Thanks for the fix!

PR #186, opened 2016-05-26
Add test for choices with duplicate display values

Ian-Foote:
Graphene fails in a not very helpful way if a django field with `choices` has duplicate display values.

I see two options here:
- forbid this, and raise a nicer exception
- allow this

I prefer the second option for consistency with django forms (which don't care). It is also potentially desirable when used with grouped choices (#185).

### DIFF
--- a/graphene/contrib/django/tests/test_converter.py
+++ b/graphene/contrib/django/tests/test_converter.py
@@ -117,6 +117,21 @@ def test_field_with_choices_convert_enum():
     assert graphene_type.__enum__.__members__['ENGLISH'].value == 'en'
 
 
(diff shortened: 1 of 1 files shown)


mjtamlyn:
I feel that the [restrictions on a GraphQL enum type](http://graphql.org/docs/api-reference-type-system/#graphqlenumtype) don't fit well with choices in all cases. In particular Django choices intends that the "internal" (first) representation is used to submit data, and the "external" (second) representation is used to render output data. Enum types use the "name" of the value for both cases, which may not be unique.

syrusakbary:
The Django integration now lives on its own package, and this issue is now fixed in `graphene-django` on `master. :)

Issue #182, opened 2016-05-25
Explicit registration of object types


mjtamlyn:
I'm not sure if the answer to this is "I'm doing it wrong" but here we go.

## Problem

So one thing I want to do is to be able to get information about the current user from the API. I have an `AnonNode` and a `UserNode` which both implement `UserInterface`. I think add a field `current_user = graphene.Field(UserInterface)` with an appropriate resolver. `UserNode` is also explicitly registered in other places, but `AnonNode` is not, I'm not using it for anything other that this specific field. If I query the field as an anonymous user, I get the following error:

```
Runtime Object type "AnonNode" is not a possible type for "UserInterface".
```

If I explicitly register the `AnonNode` by adding a meaningless field anywhere in my schema which uses that type directly, everything works.

I was hoping to be able to use `schema.register(AnonNode)`, but this seems to only update `graphene.Schema` and not the underlying `graphql.Schema`. The new type becomes present in `my_schema.type_map` but not `my_schema.schema.type_map`. The latter appears to be largely immutable.

## Possible solutions
- Deal with having to have every type associated directly with a field, but document that in the interfaces docs
- Update `schema.register` to allow explicit registration of arbitrary object types.
- Add some other API for explicit registration.
- Update `graphene.Field` to give it an extra `implementations` arg or similar when registering an `Interface` (this could  - anything matching that interface should be valid).

syrusakbary:
I think is possible to fix `schema.register` adding the type also in the `types` argument in `GraphQLSchema` (GraphQL `v0.5.0`)

If you please can post (here or creating a test scenario in a branch) the code for showcasing this bug would be great. So I can be make sure that the fix is working :)

mjtamlyn:
I think that's a minimal test case.

jameswyse:
👍  I was also having difficulty understanding/implementing this.

syrusakbary:
Just pushed a fix into master.

PR #187, opened 2016-05-28
Register interface failure fix

syrusakbary:
This PR fix the issue #182 based in the PR #183 by the work of @mjtamlyn

### DIFF
--- a/graphene/core/schema.py
+++ b/graphene/core/schema.py
@@ -77,6 +77,7 @@ class Schema(object):
             self,
             query=self.T(self.query),
             mutation=self.T(self.mutation),
+            types=[self.T(_type) for _type in list(self._types_names.values())],
             subscription=self.T(self.subscription))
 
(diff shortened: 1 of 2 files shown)


graphene/core/schema.py
```diff
@@ -77,6 +77,7 @@ def schema(self):
             self,
             query=self.T(self.query),
             mutation=self.T(self.mutation),
+            types=[self.T(_type) for _type in list(self._types_names.values())],
```
ekampf: Why do you need to cast `self._types_names.values()` to a `list`? `dict.values()` is already a list ...

syrusakbary: @ekampf in Python 3 dict.values() returns an iterable that raises an error if the dict is changed in runtime. Casting to a list fix this problem.

See the related failed testing: https://travis-ci.org/graphql-python/graphene/builds/133549359

PR #194, opened 2016-06-03
Added SimpleConnection to relay types

Globegitter:
Sometimes you do not want a full blown relay connection, where you have `edges[...].node.attribute` but simply `edges[...].attribute`. According to https://github.com/graphql/graphql-relay-js/issues/27#issuecomment-142421989 relay has support for this simpler use-case as well if you do not need certain functionality such as deletion. This implements such a type that can be used instead of the default connection via the `connection_type` argument on `ConnectionField`

This is dependent on https://github.com/graphql-python/graphql-relay-py/pull/6

### DIFF
--- a/graphene/relay/__init__.py
+++ b/graphene/relay/__init__.py
@@ -16,4 +16,4 @@ from .types import (
 from .utils import is_node
 
 __all__ = ['ConnectionField', 'NodeField', 'GlobalIDField', 'Node',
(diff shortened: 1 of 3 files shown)


Globegitter:
Also happy to add tests if that is something you feel makes sense to merge in :)

Globegitter:
Ok, abandoning that for now after disscussion at: https://github.com/facebook/relay/issues/1201

Might pick up this PR again at a later point if it turns out to be useful for us or anyone else.

PR #196, opened 2016-06-06
If node type_name is in connection type_name use connection type_name

Globegitter:
Let's say we have a node `Holdings` and then make a custom connection `HoldingConnection` that subclasses from `relay.Connection` it would then be of type `HoldingHoldingConnection`. This way we only take the connection name and don't have these redundant type names (which can become quite apparent when using relay)

### DIFF
--- a/graphene/relay/types.py
+++ b/graphene/relay/types.py
@@ -81,8 +81,11 @@ class Connection(ObjectType):
         edge_type = edge_type or Edge.for_node(node)
         assert is_node(node), 'ObjectTypes in a connection have to be Nodes'
         edges = List(edge_type, description='Information to aid in pagination.')
+        connection_name = (cls._meta.type_name
+                           if node._meta.type_name.lower() in cls._meta.type_name.lower()
+                           else '%s%s' % (node._meta.type_name, cls._meta.type_name))
         return type(
-            '%s%s' % (node._meta.type_name, cls._meta.type_name),
+            connection_name,
             (cls,),
             {'edge_type': edge_type, 'edges': edges})
 


mjtamlyn:
Should there also be a way to explicitly name any given type? This could be useful for refactoring/changing the backend

Globegitter:
Yeah, I suppose the metaclass could have an additional field `connection_type_name` or something.

mjtamlyn:
I was actually wondering if this should be "deeper" than the connection and there should be an option on the core `ObjectType`

Globegitter:
@mjtamlyn what then if the core `ObjectType` is being used in different `ConnectionFields`? Maybe it should be an option on the `ConnectionField` itself?

mjtamlyn:
Sorry, by core `ObjectType` I meant the actual `ObjectType` class, rather than the target of the connection. `Connection` is a subclass of `ObjectType`, and you can specify the `connection_type` [when creating `ConnectionField`](https://github.com/graphql-python/graphene/blob/1711e6a52907b3061f1faeca8ac629c985e28acf/graphene/relay/fields.py#L14), so that would let you name it explicitly in the case, but also explicitly name any other type in your system.

Basically we're disconnecting the python name of the class and the schema name of the type.

Globegitter:
@mjtamlyn yeah I like the idea, though I do think it should be optional as pretty much anywhere in graphene the class-name is coupled to the typename as a default.

mjtamlyn:
Of course, it's a good convention to have the names matching but I'm sure there are times when it would be inconvenient.

syrusakbary:
Makes sense! Sorry for reviewing so late, I've been spending a lot of time in the next version of graphene lately!
Please take a look and let me know your thoughts!

https://github.com/graphql-python/graphene/tree/1.0

syrusakbary:
Closing PR, seems to be directed to `graphene` `0.10.X` and it's a behavior that should be fixed in `1.X`.
(Feel free to reopen another PR if this is not fixed in `master`)

PR #201, opened 2016-06-08
Add connection_from_list_slice support to Connection

Globegitter:
This allows to return e.g. `Connection.for_node(Node).from_list([1, 2, 3, 4], args, context, info, 5389)` to get pagination support for sliced lists.

### DIFF
--- a/graphene/relay/fields.py
+++ b/graphene/relay/fields.py
@@ -29,7 +29,7 @@ class ConnectionField(Field):
     @with_context
     def resolver(self, instance, args, context, info):
         schema = info.schema.graphene_schema
-        connection_type = self.get_type(schema)
+        connection_type_for_node = self.get_type(schema)
 
         resolver = super(ConnectionField, self).resolver
         if has_context(resolver):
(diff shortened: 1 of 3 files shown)


mjtamlyn:
This looks like a good way to allow simple optimisations.

Globegitter:
Not sure why the test is failing. Will investigate tomorrow.

Globegitter:
Actually on top of that it would also be nice to be able to not pass in the count, but instead `hasNextPage` or `hasPreviousPage`. Just starting to write our pagination and some of our service will return these values directly. Just a note for myself.

Globegitter:
Also this is a good article on pagination with relay: https://m.alphasights.com/how-to-do-effortless-pagination-with-relay-connections-and-graphql-ruby-a534ffaf5cbf#.qm18yr7gm

Globegitter:
We have now actually resorted to create the connection and edges ourselves like so:

```
def _create_statement_edge(statement):
    edge_type = relay.Edge.for_node(StatementEntry)
    return edge_type(node=statement, cursor=base64.b64encode(str(statement.id)))

...
def resolve_bla(...):
    pagination_param = base64.b64decode(args['after']) if args.get('after') else None
    
    statements, has_next_page  = call_to_database_service(pagination_param)

    edges = [_create_statement_edge(statement) for statement in statements]
    connection_type = relay.Connection.for_node(StatementEntry)

    return connection_type(edges=edges, page_info=relay.PageInfo(has_next_page=has_next_page))
```

There is a bug though with returning `connection_type` in this way, which I fixed here: https://github.com/graphql-python/graphene/pull/237

PR #204, opened 2016-06-13
fixed model column name [sqlalchemy]

pyoner:
I have existing db and create User model for that db using automap_base
```python
import os

from sqlalchemy.ext.automap import automap_base
from sqlalchemy.orm import Session
from sqlalchemy import create_engine, Column, Integer, String, ForeignKey

Base = automap_base()
class User(Base):
    __tablename__ = 'users'
    id = Column(Integer, primary_key=True)
    email = Column('mail', String(100))
    password = Column('pass', String(32))

engine = create_engine(os.getenv('DATABASE_URI'))
# reflect the tables
Base.prepare(engine, reflect=True)
```

but if I use query for users
```
query {
  users {
    edges {
      node {
        about
        email
        password
      }
    }
  }
}
```
fields `email` and `password` doesn't exists because these fields are redefined in the model


syrusakbary:
Makes sense! Thanks for the fix :)

PR #208, opened 2016-06-21
Fix typo: CategoryNode -> PostNode

ianlini:
There is no `CategoryNode` in the authorization docs page and I believe it's actually `PostNode`,
so I fixed 2 words in the docs: `CategoryNode` -> `PostNode`


syrusakbary:
Thanks for the catch!

PR #210, opened 2016-06-22
[Docs] Make it more obvious that the `node` Field is required for relay

reisub-de:
For me it was not obvious that you need a node Field in your Query. I hope this makes the documentation a bit better. Also it now gives the error message that occurs if you forget the node Field, so googling for this brings you right to the documentation.

It took me way longer than I'd like to admit to find this ^^


syrusakbary:
Thanks! Merging

PR #213, opened 2016-06-25
Added returns_promise flag to schema.execute

AlecAivazis:
This PR adds an additional flag to the `execute` method of the schema so that the execution returns a promise.


syrusakbary:
There is a bug in the pytest coverage package in Python 3.5... that's why tests failed, your changes are perfectly valid.
Merging!

PR #217, opened 2016-07-01
Wrapped enum type in scalar types

### LINKED ISSUE #205: Django Model charField with choices : query field
I have some problems when I try to query a Django charField with choices.

I have the following models.py :

``` python
class Vehicle(models.Model):
    TYPE_CHOICES = (
        ('MOTO', _('MOTORBIKE')),
        ('CAR', _('car')),
        ('BIKE', _('bike')),
    )

    name = models.CharField(max_length=100)

    vtype = models.CharField(
        _('Vehicle type'),
        max_length=5,
        choices=TYPE_CHOICES,
        default='MOTO',
        help_text=_('Lorem ipsum')
    )

    description = models.TextField(blank=True, default='')

    def __str__(self):
        return self.name
```

and the following schema.py :

``` python
  class VehicleNode(DjangoNode):
      class Meta:
          model = Vehicle
```

When I try to make this query :

``` javascript
query {
  allVehicles {
    edges {
      node {
        id,
        name,
        vtype
      }
    }
  }
}
```

I have the following error :

``` javascript
{
  "errors": [
    {
      "message": "Cannot query field \"vtype\" on type \"VehicleNode\".",
      "locations": [
        {
          "line": 7,
          "column": 9
        }
      ]
    }
  ]
}
```

To be able to query vtype I need to add `vtype  = String()` into my schema like that :

``` javascript
class VehicleNode(DjangoNode):
    vtype  = String()
    class Meta:
        model = Vehicle
```

I don't know if I missed something ?

sximba:
This fixes  #205 

I have not been able to write formal tests for it yet. Could you please have a look at it and let me know if I am on the right track and then I will add the tests.


syrusakbary:
This is now fixed in `graphene-django` in master. Closing :)

PR #235, opened 2016-07-26
Allow setting of any attributes starting with _ (private attributes)

### LINKED ISSUE #234: Instantiate graphql types with _ attributes
Given a type e.g.

```
class Some(graphene.ObjectType):
  _state = None // some state I do not want to have in the schema
 _other_state = None
```

I have to instantiate this as follows which is quite cumbersome and feels very unpythonic:
```
s = Some()
s._state = 'secret'
s._other_state = 'other_secret'
```
where-as it would be nice to be able to just do `s = Some(_state='secret', _other_state='other_secret')`. I know you can do `s = Some(_root={'_state': 'secret', '_other_state': 'other_secret'})` but that makes it far less explicit about what state I am sharing and especially in a big schema this can turn into a problem quickly as it is very easy to loose track of what is shared with what type.

So it would be nice if we could get the best of both worlds and just allow `s = Some(_state='secret', _other_state='other_secret')

Globegitter:
This fixes #234


Globegitter:
Not sure while the build fails on py 3.5 but it seems unrelated to my changes.

Globegitter:
@syrusakbary Any word on this PR? Is that something you could see getting merged in? Would allow to clean up our source code a bit.

PR #237, opened 2016-07-26
Allow ConnectionFields to have ObjectTypes as per relay spec

Globegitter:
In the Relay spec it says that the `node` of a `ConnectionField` does not actually have to be a node but can also be an objectType: https://facebook.github.io/relay/graphql/connections.htm#sec-Node

### DIFF
--- a/examples/starwars_relay/tests/test_connections.py
+++ b/examples/starwars_relay/tests/test_connections.py
@@ -34,8 +34,5 @@ def test_correct_fetch_first_ship_rebels():
         }
     }
     result = schema.execute(query)
-    print('-------------------------------')
-    print(result)
-    print(result.errors)
     assert not result.errors
     assert result.data == expected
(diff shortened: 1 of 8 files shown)


Globegitter:
Also as a note, we are using this patch already locally in our graphene. So really hope this functionality makes it in. Even if it is in a different way with 1.0.

I am also happy btw to make these last PRs agains the 1.0 branch if you want @syrusakbary

Globegitter:
I know my last commit is not directly related but since I am already touching the code I thought I'd fix that here as well. This is something we notice when trying to return a Connection directly instead of a list, that it actually does not work with the given `isinstance` check.

syrusakbary:
Good catch! I would like to merge.

I'm refactoring the base code of graphene, so would be great if we have tests covering the use case so we assure support for this in the future (the tests will remain very similar even in the refactor).

Globegitter:
@syrusakbary Added tests for both fixes in here :)

PR #238, opened 2016-08-04
Convert `orderBy` value to snake case

jkimbo:
I came across then when trying to use `orderBy` on the `DjangoFilterConnectionField` and found I had to use the snake case version of the field name for it to actually filter values when I would have expected to be able to use the camel case version. This PR allows you to use camel case for the field name in `orderBy`.

Not sure if this breaks any current assumptions but the tests seem to pass!

### DIFF
--- a/graphene/contrib/django/filter/fields.py
+++ b/graphene/contrib/django/filter/fields.py
@@ -1,3 +1,4 @@
+from ....utils import to_snake_case
 from ..fields import DjangoConnectionField
 from .utils import get_filtering_args_from_filterset, get_filterset_class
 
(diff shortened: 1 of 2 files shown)


graphene/contrib/django/filter/fields.py
```diff
@@ -26,7 +27,13 @@ def get_queryset(self, qs, args, info):
         filter_kwargs = self.get_filter_kwargs(args)
         order = self.get_order(args)
         if order:
-            qs = qs.order_by(order)
+            qs = qs.order_by(to_snake_case(order))
```
slorg1: should `get_order` which is the function designed to handle the API level (`args`) to the `to_snake_case` so it would "own" the API layer and let `get_queryset` deal with `qs` level things? (separation of concerns / layering).

slorg1: Other question: what if `order` is a list? https://docs.djangoproject.com/en/1.10/ref/models/querysets/#django.db.models.query.QuerySet.order_by

To do multi field ordering (in your test `order_by('first_name', 'articles')`)

`django_filters.filterset.BaseFilterSet`
```                    qs = qs.order_by(*self.get_order_by(ordered_value))
```
would handle it, though (since I do not need a nice patch from you for this one) it would not handle the nice `to_snake_case` you added.

jkimbo: @slorg1 you're probably right about moving the snake case translation into `get_order` as that is probably cleaner. I also don't think graphene currently supports ordering by multiple fields. Possibly raise that as a separate issue/PR with a failing test case?

PR #241, opened 2016-08-05
Added --indent option for graphql_schema command

jkimbo:
I'll create another PR for the docs (btw I'm not sure the reasoning around splitting the docs out into a separate repo but it makes things like this quite inconvenient!)

### DIFF
--- a/graphene/contrib/django/management/commands/graphql_schema.py
+++ b/graphene/contrib/django/management/commands/graphql_schema.py
@@ -25,6 +25,13 @@ if LT_DJANGO_1_8:
                 default='',
                 help='Output file (default: schema.json)'
             ),
+            make_option(
+                '--indent',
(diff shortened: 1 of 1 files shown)


syrusakbary:
Thanks for the suggestion, It might be possible that we merge the docs back.
The main reason for keep them separate was isolating the docs/website building away from the travis testing, but there might be a intermediate formula to make all people happy! :)

jkimbo:
My personal preference is keeping docs with the code since it means that there is less chance of them getting out of sync (maintaining docs is hard enough already!).

Anyway regardless I didn't mean to come across as ungrateful or anything! It's a fantastic project and I'm enjoying using it. Got some time on my hands at the moment as well so I'll try and pick off some of the low hanging issues if thats ok?

Issue #66, opened 2015-12-09
Share fields between mutation input and relay Node


Globegitter:
[...]

PR #247, opened 2016-08-16
Added breaking test to demonstrate wanted subclassing feature.

Globegitter:
This is a feature that would really help us clean up our code without affecting the schema. In my opinion putting the `interfaces` on  the meta class is a great step into the right direction already and should simplify this step.

I will also add a test for subclassing/sharing fields between mutation inputs types and normal types (#66). Will also have a bit of a dig into the code-base and see if I can implement that myself.

### DIFF
--- a/graphene/relay/tests/test_mutation.py
+++ b/graphene/relay/tests/test_mutation.py
@@ -27,10 +27,13 @@ class OtherMutation(ClientIDMutation):
     class Input(SharedFields):
         additional_field = String()
 
+    name = String()
+
     @classmethod
     def mutate_and_get_payload(cls, args, context, info):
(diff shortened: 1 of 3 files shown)


Globegitter:
I only added the tests to relay right now as this is what we are using, but happy to add them to standard graphql as well if accepted.

Globegitter:
Ahh, I have just found the `AbstractType`. Great addition! the only thing I am noticing is that the subclass does not inherit the interface that the abstract type implements. Would be nice to get that feature as well.

syrusakbary:
Let me reopen this PR, I think makes sense to have `AbstractTypes` inheritance in Mutation Inputs.

PR #248, opened 2016-08-16
Inhert abstracttype interface

Globegitter:
This supersedes #247 to add functionality to the new `1.0` version to also inherit the interfaces (and their field definitions) if an `ObjectType` inherits from an `AbstractType`. Added a quick test and implementation to get it to work.

Will clean up asap (probably in two days time as I have tomorrow off) and add some more tests. Let me know what you think though.

Looking forward to `1.0`. Thanks for all the work!

### DIFF
--- a/graphene/types/abstracttype.py
+++ b/graphene/types/abstracttype.py
@@ -1,3 +1,4 @@
+from collections import OrderedDict
 import six
 
 from ..utils.is_base_type import is_base_type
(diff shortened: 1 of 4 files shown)


syrusakbary:
Hi @Globegitter, thanks for the contribution!

`AbstractTypes` are meant to be generic and and allow inheritance to `ObjectTypes`, `Interfaces` and `InputObjectTypes`.
However the specific config should live in the corresponding type. `AbstractType` it's only meant to be used for fields (which is common among all the types that I described before), but other things like `interfaces` are too specific for living in a "global" type.

Globegitter:
@syrusakbary we have a use case though where `AbstractType` is perfect but we also need to be able to guarantee that the separate `ObectTypes` inheriting from this `AbstractType` implement the same `Interface`. This PR enables this use-case. I can see though why you might not want that. Would there be a possibility to add a more specific `AbstractObjectType` to allow for our specific use case? Or otherwise the `AbstractType` could only pass on the `Interface` specification to `ObjectTypes` but nothing else. Or any other ideaa to allow for this?

Globegitter:
@syrusakbary Is there any support for this in `1.0`, or what was the conclusion you came to? This is still something quite important to us.

syrusakbary:
Hi Markus,

I closed the PR by mistake, I think we should still try to find a way for make it work.
If you wanna reopen the PR and discuss further I'm happy with that! :)

Globegitter:
@syrusakbary Thanks, will add a new PR soon :)

PR #253, opened 2016-08-19
Fix return promise from ConnectionField resolver

Globegitter:
This is a quick-fix so we can get `ConnectionField` resolvers work right now in our app when returning promises.

Will add a test and I think in your `next` branch we can make a nicer fix to actually use the `promise` `is_thenable` check.


Globegitter:
Not sure why the tests are failing. Are working locally for me.

Globegitter:
This is being fixed in https://github.com/graphql-python/graphene/pull/288.

PR #266, opened 2016-08-31
next: handle objects being constructed in the typemap multiple times

sjhewitt:
When using a Union type, it can happen that `construct_objecttype` is called multiple times for the same type. This change bails early if the object is already in the map.

### DIFF
--- a/graphene/types/typemap.py
+++ b/graphene/types/typemap.py
@@ -119,6 +119,8 @@ class TypeMap(GraphQLTypeMap):
     @classmethod
     def construct_objecttype(cls, map, type):
         from .definitions import GrapheneObjectType
+        if type._meta.name in map:
+            return map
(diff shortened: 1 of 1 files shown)


graphene/types/typemap.py
```diff
@@ -119,6 +119,8 @@ def construct_enum(cls, map, type):
     @classmethod
     def construct_objecttype(cls, map, type):
         from .definitions import GrapheneObjectType
+        if type._meta.name in map:
```
syrusakbary: For consistency we should also check if `map[type._meta.name].graphene_type == type` :)

syrusakbary:
I closed this PR by mistake. Will merge your changes! :)

syrusakbary:
Merged with https://github.com/graphql-python/graphene/commit/061c33f8d611f6a3bd0d69cebbf220f7cda4955f

PR #281, opened 2016-09-12
Export union type directly.

Globegitter:
Exports union type and fixes a missing comma.


Globegitter:
Also added `GlobalID` to be exported.

PR #282, opened 2016-09-12
Making node argument optional for GlobalID and add clientMutationId to Relay mutation response.

Globegitter:
First of all, love that `GlobalID` now takes a `node` argument, will allow us using it for Mutations, which is great. But we often also have the case where we implement a `Node` interface in an `ObjectType` but want to define `id = some_annotation_function(GlobalID())` on that `ObjectType`. This PR allows for that case.


Globegitter:
Sorry btw, for this overlap in functionality change in this PR. But it is still a small diff so should hopefully be fine. But let me know if you want me to split this out in 2 PRs.

syrusakbary:
I think the `clientMutationId` fix is good to go!

However I'm not sure if I see the use case of decorating the `GlobalID`. Instead of returning `node_id` maybe is better to return `node` (a field referencing the `Node` interface) and get `node { id }`. Thoughts?

Globegitter:
In some cases, e.g. a `NODE_DELETE` relay expects the global id of a thing to be sent back directly rather than nested (see https://facebook.github.io/relay/docs/guides-mutations.html#node-delete). And I have always found it would be nice if we could specify that in the schema rather than how we do it currently: specifying `graphene.ID()` and then using `to_global_id(NodeType._meta.typename, 'local_id')`.

Globegitter:
@syrusakbary what do you think of this given that relay sometimes expects a global id/global ids to be returned directly?

Globegitter:
@syrusakbary just started upgrading our codebase to `1.0` again and the fact that `relay.GlobalID()` needs a `node` parameter is starting to clutter up our codebase. What is blocking this PR from getting merged in?

Globegitter:
Just stumbling over more things again. To describe our use-case further, whenever we have an objecttype that implements `relay.Node` we re-define `relay.GlobalID()` directly on that type because we need to add some metadata specific to our system (the annotation function mentioned above).

So whenver we have an objectType we would have:

```
class SomeClass(graphene.ObjectType):

    class Meta:
        interfaces = (relay.Node, )

    id = annotation_function(relay.GlobalID(relay.Node, required=True), 'some_metadata')
```

But the syntax around `GlobalID` in `0.x` seemed much cleaner to me and it would just be:
```
class SomeClass(graphene.ObjectType):

    class Meta:
        interfaces = (relay.Node, )

    id = annotation_function(relay.GlobalID(), 'some_metadata')
```

The above syntax would still be valid, just that it now has more sensible defaults.

PR #283, opened 2016-09-12
Re-add default value for Field.

Globegitter:
This used to exist in `0.x` (with `default`) though. We are using it in a few cases and I do think it is a quite nice sugar.

Also this time it does not allow `callable` since that doesn't really seem to bring any benefits anymore over just providing the `resolver` function.

### DIFF
--- a/graphene/types/field.py
+++ b/graphene/types/field.py
@@ -2,6 +2,8 @@ import inspect
 from collections import Mapping, OrderedDict
 from functools import partial
 
+from promise import Promise
+
 from ..utils.orderedtype import OrderedType
 from .argument import to_arguments
 from .structures import NonNull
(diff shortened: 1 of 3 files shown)


syrusakbary:
I think it could be useful to add `default_value` back into `Field`, however I think it only should be applied when using the default resolver, as if not it could have side-effects if the custom resolver function returns `None`.

For that, we can do two things:
* Add `defaut_value` into Field, and modifying the [Typemap default resolver](https://github.com/graphql-python/graphene/blob/master/graphene/types/typemap.py#L189) to use the default value if the attr is not found in root. Similar to `getattr(root, attname, default_value)`
* Create a resolver decorator that mimics the logic that you did.

If the latter is the best approach to make, maybe keeping the decorator in your app scope is a good case until more people have the same issue.
Let me know your thoughts!

Globegitter:
I'll have a bit of a think about it but I like the sound of the first solution :)

Globegitter:
New PR: https://github.com/graphql-python/graphene/pull/298

PR #288, opened 2016-09-14
Reworks of relationship between Nodes/Objects, Connections and Edges

Globegitter:
…Edges.

I like a lot of the stuff that is happening on 1.0 :). After having looked more into how the relationship of connections to object types has been reworked I think there is a more accurate way of modeling their relationship.

To my understanding a `Node` or an `ObjectType` can belong to many `Connections`. A Connection has only one `Edge` (where-as an `Edge` can belong to many Connection) and an `Edge` has only one `Node/ObjectType` (where-as again a `Node/ObjectType` can belong to many `Edges`).

This should allow for any of these use-cases by removing `NodeType.Connection`, since it seems to me a `Node` should not now anything about a `Connection`. Furthermore this also leaves out a normal `ObjectType` (which can be used for a `Connection`).

Further I added a `Connection.for_type(GqlType)` so we can get the automatically created `ConnectionType`. This uses a thread-safe caching library based on the python 3 `lru_cache` api, I assume the `memoize` might have been part of the multi-threading issues. If we don't want to use memoizing at all the only other ways I can see are:
- somehow getting the auto-created connection from the schema's `TypeMap`
- storing the auto-created connection on a dict on the `ObjectType`/`Node` and then providing a function to retrieve this ala `get_auto_generated_connection_from_type(GqlType, 'ConnectionName')` but that kind of seems a bit sucky.
- somehow getting it from the `ConnectionField` which is responsible for auto-creating the connection
- disabling auto-creation from `Connections` completely (i.e. disallowing creating a `ConnectionField with a`Node/ObjectType`) and just making it easy to create your own`Connection`by providing`AbstractTypes`and maybe`Connection.for_type`function that, which then creates the connection that is used for the`ConnectionField`

Having thought about this a bit more, I am actually quite in favor of the last option, which would only require minor changes to this PR.

Anyway, I know this is a lot of text and not sure if that is all clear, but there should be tests for all changes, which might make it easier to understand. Let me know what you think @syrusakbary.

Btw, apart from the rework this also includes the fixes from https://github.com/graphql-python/graphene/pull/253 and allows for a resolver to return a `Connection` directly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -75,6 +75,7 @@ target/
 
 # PyCharm
 .idea
+*.iml
 
 # Databases
 *.sqlite3
--- a/examples/starwars_relay/schema.py
+++ b/examples/starwars_relay/schema.py
@@ -16,6 +16,8 @@ class Ship(graphene.ObjectType):
     def get_node(cls, id, context, info):
         return get_ship(id)
 
+ShipConnection = relay.Connection.for_type(Ship)
+
 
 class Faction(graphene.ObjectType):
     '''A faction in the Star Wars saga'''
@@ -24,7 +26,7 @@ class Faction(graphene.ObjectType):
         interfaces = (relay.Node, )
 
     name = graphene.String(description='The name of the faction.')
-    ships = relay.ConnectionField(Ship, description='The ships used by the faction.')
+    ships = relay.ConnectionField(ShipConnection, description='The ships used by the faction.')
 
     @resolve_only_args
     def resolve_ships(self, **args):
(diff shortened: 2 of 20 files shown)


Globegitter:
Ok, I am also seeing that `fastcache` has some issues running on `pypy`. I could investigate that or switching to a thread-safe `memoize`, but I am leaning more and more towards `ConnectionField` only taking a `Connection` and putting the responsibility to the user rather than the framework trying to be a bit too clever.

Globegitter:
I now implemented what I proposed - removing the 'magic' auto generation of `Connections` in favour of declarative genaration of `Connections`, which should remove any issues of caching as well as tying any Connections to a `Node` but also allowing `ObjectTypes` to be used for `Connections` as specified in the Relay spec.

Globegitter:
Can be closed in favour of #347

PR #295, opened 2016-09-19
Object type string and repr method

Globegitter:
Seems to me a nice thing to have. The main question though is, would we also want to show empty fields or any information about the resolvers? (Conceptually I see an ObjectType not just as an object that has data but an object that has data as well as knowledge on how to fetch data)


Globegitter:
Not sure why `pypy` is failing with:

```
ImportError while importing test module '/home/travis/build/graphql-python/graphene/graphene/types/tests/test_objecttype.py'.
Original error message:
'No module named mock'
Make sure your test modules/packages have valid Python names.
```

Any ideas?

PR #297, opened 2016-09-19
relay Connection.node does not have to subclass Node/ObjectType

sjhewitt:
https://facebook.github.io/relay/graphql/connections.htm#sec-Node


syrusakbary:
Good catch! 👍

PR #298, opened 2016-09-21
Added default value for default resolver.

Globegitter:
This is implementing `default_value` as you suggested @syrusakbary in https://github.com/graphql-python/graphene/pull/283.


khankuan:
@syrusakbary Nice! Is there an example on using the default resolvers?

PR #299, opened 2016-09-21
Rename type to gql_type in Field to fix failing test in master.

Globegitter:
Not sure how merging this into master changed this back, but here is the fix to get the tests passing again (Yeah it is really annoying for graphql that `type` is a python builtin)

### DIFF
--- a/graphene/types/field.py
+++ b/graphene/types/field.py
@@ -17,7 +17,7 @@ def source_resolver(source, root, args, context, info):
 
 class Field(OrderedType):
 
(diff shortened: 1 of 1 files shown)


syrusakbary:
Hey Markus,

I amended the merge commit changing from `gql_type` to `type`.
The reason for that is trying to preserve compatibility as I'm aware of some cases where developers where using `Field(type=XXX, ...)`.

Now should be fixed on master. Thanks for the PR though!

Globegitter:
Oh yes, good point did completely not think about this.

PR #301, opened 2016-09-22
fix the hello world example in the README

adamhadani:
After initially checking the project out and trying to use it I ran into an error when using the example schema from the README as is. After looking through the source code and also comparing it to the version I saw in the website's getting started docs (http://graphene-python.org/docs/quickstart/), seems like it's a simple fix..

As an aside, perhaps it would be good to get rid of one of the README's so easier to maintain going forward? seems like there's both .md and a .rst versions currently, I updated both for good measure..

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ Here is one example for get you started:
 class Query(graphene.ObjectType):
     hello = graphene.String(description='A typical hello world')
 
-    def resolve_hello(self, args, context, info):
+    def resolve_hello(self, args, info):
         return 'World'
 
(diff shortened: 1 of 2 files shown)


syrusakbary:
Hi @adamhadani, the newest version of graphene (`1.0.dev`, that is on `master`) take now always the context argument.
Probably you are using graphene 0.x, (you can check the README in the 0.x branch, it should be similar to the one you pointed)

adamhadani:
got it @syrusakbary - thanks for clarifying! makes sense

PR #310, opened 2016-10-03
Added two more fields to convert.py

Momotv45:
Added:
* PhoneNumberField
* DurationField

Both of which come with Django!


syrusakbary:
Hi @Momotv45, I guess you forked graphene some time ago as now the `core` has been rebuilt and the Django integration is living in its own repo: https://github.com/graphql-python/graphene-django/

I'm closing here the PR but feel free to open it in `graphene-django`.

Thanks anyway for your work!

PR #318, opened 2016-10-07
Fixed typos and improved language and markup

ksonj:
I tried my best to fix some aesthetic issues with the documentation that I stumbled upon. As English is not my first language either I tried to be very conservative, although I feel like the wording of the "Getting Started" guide could still be further improved.


syrusakbary:
Thanks for contributing and sorry about taking so long for merging!

PR #325, opened 2016-10-13
Allow setting unmounted types with custom attributes that get carried…

Globegitter:
… over.

We are often putting metadata on our Fields and with `1.0` we would not be able to use `graphene.Boolean` etc but would always have to wrap them in `graphene.Field(..)`. It would be ideal if this could get merged in to save us this wrapping in cases where we need this metadata, making the API consistent across our whole apps. (See the tests for some made up but close enough use-cases)

### DIFF
--- a/graphene/types/tests/test_mutation.py
+++ b/graphene/types/tests/test_mutation.py
@@ -44,15 +44,30 @@ def test_mutation_raises_exception_if_no_mutate():
 
 
 def test_mutation_execution():
+    def with_metadata(gql_type, data):
+        setattr(gql_type, 'metadata', data)
+        return gql_type
+
     class CreateUser(Mutation):
         class Input:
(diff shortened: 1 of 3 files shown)


syrusakbary:
Hi Marcus,

I think this PR is very tied to the application logic instead of the framework... I'm not sure how this could be useful to other developers.

Maybe the best solution could be subclassing `Field` and use your custom classes with the extra attrs you need.

Globegitter:
Well, I guess my rationale behind this is: We are doing a magic conversion of an unmounted type to a field. So we might as well magically copy all attributes from the unmounted type. But maybe there is a better/easier way of doing this.

But I can also understand that decision and I am sure for this we can find a solution on application level.

Globegitter:
Replaced with #332

PR #332, opened 2016-10-21
Make scalar get_type an instance method.

Globegitter:
Given the current way of how Scalars are mostly being created in the schema, e.g. `graphene.Boolean(description='Some boolean')` and then how they get 'magically' turned into a `Field` using `self.get_type()` it makes sense to make this an Instance method rather than a classmethod. This allows e.g. to set attributes on the instance that will always be accessible, as they are being returned by `get_type`.

This replaces https://github.com/graphql-python/graphene/pull/325

Further, if you are worried about backwards-compatibility we could also keep this a class method and then simply override that classmethod with an instancemethod on instantiation. Being a 100% backwards-compatible and allowing for the use-case described,

``` py
import types
...
class Scalar(six.with_metaclass(ScalarTypeMeta, UnmountedType)):
    '''
    Scalar Type Definition
    The leaf values of any request and input values to arguments are
    Scalars (or Enums) and are defined with a name and a series of functions
    used to parse input from ast or variables and to ensure validity.
    '''

     def  __init__(self):
        def get_type(self):
            return self
        self.get_type = types.MethodType(get_type, self)

    serialize = None
    parse_value = None
    parse_literal = None

    @classmethod
    def get_type(cls):
        return cls
```

### DIFF
--- a/graphene/types/unmountedtype.py
+++ b/graphene/types/unmountedtype.py
@@ -63,7 +63,8 @@ class UnmountedType(OrderedType):
         return (
             self is other or (
                 isinstance(other, UnmountedType) and
-                self.get_type() == other.get_type() and
+                self.get_type()._meta == other.get_type()._meta and
+                # self.get_type()._meta.name == other.get_type()._meta.name and
                 self.args == other.args and
                 self.kwargs == other.kwargs
             )
--- a/graphene/relay/tests/test_connection.py
+++ b/graphene/relay/tests/test_connection.py
@@ -68,7 +68,7 @@ def test_edge():
     assert edge_fields['node'].type == MyObject
 
     assert isinstance(edge_fields['other'], Field)
-    assert edge_fields['other'].type == String
+    assert edge_fields['other'].type.__class__ == String
 
 
 def test_edge_with_bases():
@@ -92,7 +92,7 @@ def test_edge_with_bases():
     assert edge_fields['node'].type == MyObject
 
     assert isinstance(edge_fields['other'], Field)
-    assert edge_fields['other'].type == String
+    assert edge_fields['other'].type.__class__ == String
 
 
 def test_edge_on_node():
--- a/graphene/relay/tests/test_mutation.py
+++ b/graphene/relay/tests/test_mutation.py
@@ -81,7 +81,7 @@ def test_mutation():
     assert field.args['input'].type.of_type == SaySomething.Input
     assert isinstance(fields['client_mutation_id'], Field)
     assert fields['client_mutation_id'].name == 'clientMutationId'
-    assert fields['client_mutation_id'].type == String
+    assert fields['client_mutation_id'].type.__class__ == String
 
 
 def test_mutation_input():
@@ -90,9 +90,9 @@ def test_mutation_input():
     fields = Input._meta.fields
     assert list(fields.keys()) == ['what', 'client_mutation_id']
     assert isinstance(fields['what'], InputField)
-    assert fields['what'].type == String
+    assert fields['what'].type.__class__ == String
     assert isinstance(fields['client_mutation_id'], InputField)
-    assert fields['client_mutation_id'].type == String
+    assert fields['client_mutation_id'].type.__class__ == String
 
 
 def test_subclassed_mutation():
@@ -113,11 +113,11 @@ def test_subclassed_mutation_input():
     fields = Input._meta.fields
     assert list(fields.keys()) == ['shared', 'additional_field', 'client_mutation_id']
     assert isinstance(fields['shared'], InputField)
-    assert fields['shared'].type == String
+    assert fields['shared'].type.__class__ == String
     assert isinstance(fields['additional_field'], InputField)
-    assert fields['additional_field'].type == String
+    assert fields['additional_field'].type.__class__ == String
     assert isinstance(fields['client_mutation_id'], InputField)
-    assert fields['client_mutation_id'].type == String
+    assert fields['client_mutation_id'].type.__class__ == String
 
 
 # def test_node_query():
--- a/graphene/types/scalars.py
+++ b/graphene/types/scalars.py
@@ -41,9 +41,8 @@ class Scalar(six.with_metaclass(ScalarTypeMeta, UnmountedType)):
     parse_value = None
     parse_literal = None
 
-    @classmethod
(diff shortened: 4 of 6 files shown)


Globegitter:
Ok, it was not quite as simple as I initially thought but was not that much more effort. I did it now in the slightly more backwards-compatible way. Let me know what you think about this approach @syrusakbary

graphene/types/unmountedtype.py
```diff
@@ -63,7 +63,8 @@ def __eq__(self, other):
         return (
             self is other or (
                 isinstance(other, UnmountedType) and
-                self.get_type() == other.get_type() and
```
Globegitter: This is the only line of this PR that I am not sure about. Since `get_type()` now returns self, this equality check ends up being infinitely recursive. But as far as I can tell all we want to compare is their graphql type information so that should be `self.get_type()._meta == other.get_type()._meta`.

Globegitter: I wasn't a 100% sure though if equality on the meta class like that just works of if we would need to compare the props directly on the meta class.

Globegitter:
Also just starting to test this patch out in our app and it does resolve our issue, while being imo far less intrusive to graphene itself.

syrusakbary:
Hi Marcus,

Is there any reason on why subclassing `Scalar` and replacing there the `get_type` function as bounded method is not a good solution?

```python
class MyCustomString(Scalar):
        super(MyCustomString, self).__init__(*args, **kwargs)
        self._meta = {'my_meta': True}
    
    def get_type(self):
        return graphene.String
```

Globegitter:
Yes we could do something like you proposed (though I am not sure if this exact snippet would work) but I am quite reluctant to that as the change between 0.x and 1.0 was an undocumented quite significant behavioural change and more importantly it would mean all of our apps would have to use our own scalars rather than the default ones. Even though most of it could be done automatically that would require a huge code change (the number of scalar fields we have probably ranges in the 1000s for our codebase) and further it would require us to ensure that this custom code still works in the future, etc.

What is the issue with having `get_type` as an instance method on `Scalar`?  It is in an instance method on `unmountedtype` so even seems to me more consistent for all subclasses to be an instance method as well.

syrusakbary:
After reviewing, I'm not super sure that adding scalar instances to the typemap conversion is a good idea.

If we do it, things like the following could be possible:
```python
class MyObjectType(graphene.ObjectType):
    my_string = graphene.Field(graphene.String())
    mg_string = graphene.Field(graphene.String) # This is the only way that works now
```
(Note that the PR will introduce two ways for doing the same thing).

I'm favorable of changing the `Scalar` implementation to something like the following, though:

```python
class Scalar(six.with_metaclass(ScalarTypeMeta, UnmountedType)):
    '''
    Scalar Type Definition

    The leaf values of any request and input values to arguments are
    Scalars (or Enums) and are defined with a name and a series of functions
    used to parse input from ast or variables and to ensure validity.
    '''

    serialize = None
    parse_value = None
    parse_literal = None

    def get_type(self):
        '''
        This function is called when the unmounted type (Scalar instance)
        is mounted (as a Field, InputField or Argument)
        '''
        return type(self)
```

Would that work?

Globegitter:
@syrusakbary That doesn't work unfortunately either. I have thought about that a bit more and would you be open to introducing some kind of metadata concept that we can store on the field definition? It would not be to difficult to implement and you would simply have a `set_metadata` method that would set a `_metadata` attribute and then we can just make sure to carry around this `_metadata` attribute. This would then actually make it a feature other people could use as well.

To give an example I could have:

```py
def set_metadata(gql_type, value):
    gql_type.set_metadata(value)
    return gql_type

class MyObjectType(graphene.ObjectType):
    my_string = set_metadata(graphene.String(), 'some_value')
    mg_string = set_metadata(graphene.Field(myScalar), 'some_other_value')
```

and internally we would have:
```py
class UnmountedType(OrderedType):
    '''
    This class acts a proxy for a Graphene Type, so it can be mounted
    dynamically as Field, InputField or Argument.

    Instead of writing
    >>> class MyObjectType(ObjectType):
    >>>     my_field = Field(String(), description='Description here')

    It let you write
    >>> class MyObjectType(ObjectType):
    >>>     my_field = String(description='Description here')
    '''

        super(UnmountedType, self).__init__()
        self.args = args
        self.kwargs = kwargs
        self._metadata = None

    def get_type(self):
        raise NotImplementedError("get_type not implemented in {}".format(self))

    def set_metadata(value)
        self._metadata = value

    def Field(self):  # noqa: N802
        '''
        Mount the UnmountedType as Field
        '''
        from .field import Field
        f = Field(
            self.get_type(),
            *self.args,
            _creation_counter=self.creation_counter,
            **self.kwargs
        )
        f.set_metadata(self._metadata)
        return f

    def InputField(self):  # noqa: N802
        '''
        Mount the UnmountedType as InputField
        '''
        from .inputfield import InputField
        f = InputField(
            self.get_type(),
            *self.args,
            _creation_counter=self.creation_counter,
            **self.kwargs
        )
        f.set_metadata(self._metadata)
        return f

    def Argument(self):  # noqa: N802
        '''
        Mount the UnmountedType as Argument
        '''
        from .argument import Argument
        a = Argument(
            self.get_type(),
            *self.args,
            _creation_counter=self.creation_counter,
            **self.kwargs
        )
        a.set_metadata(self_metadata)
        return a
```

And `Argument`, `Field` and `InputField` would of course implement these methods accordingly. We could even allow it being set at instantiation to keep it a one-liner. What do you think? Imo a worthy feature to add and will keep everything else exactly as it currently is.

Globegitter:
Or after having thought about it longer, we could even make it 'less intrusive' and not have this `set_metadata` interface and just allow for  `_metadata` to be passed in at instantiation (similar to `_creation_counter`) or otherwise we could set it directly `gql_type._metadata = 'dfs'` and that just gets on passed through e.g.

```
def Field(self):  # noqa: N802
        '''
        Mount the UnmountedType as Field
        '''
        from .field import Field
        return Field(
            self.get_type(),
            *self.args,
            _creation_counter=self.creation_counter,
            _metadata=self._metadata,
            **self.kwargs
        )
```

This is imo less pythonic but as I said would make it less intrusive.

syrusakbary:
Closing this PR in favor of #392 :)

Issue #341, opened 2016-10-27
ConnectionBase.edges should be NonNull


barakcoh:
We're using `gql2flow` to generate Flow types from the GraphQL schema and found that all of our `Connection` fields are defined with `edges` being nullable. 

According to [`graph-relay-js`](https://github.com/graphql/graphql-relay-js/blob/5927f0becf43740a1f1026c6e960811c4359d342/src/connection/connectiontypes.js) seems that `ConnectionBase.edges` should be defined as `NonNull(List(edge))`

Is there a scenario in which `edges` may be `None`?

syrusakbary:
This should be fixed with #342 (thanks to @ekampf work).

PR #342, opened 2016-10-27
Connection Edges should be NonNull

ekampf:
Ass defined in the reference implementation on https://github.com/graphql/graphql-relay-js/blob/5927f0becf43740a1f1026c6e960811c4359d342/src/connection/connectiontypes.js

Solves issue #341


syrusakbary:
Thanks Eran!

Issue #304, opened 2016-09-28
Implementation of Connections somewhat broken in 1.0


Globegitter:
First of all, excited to see `1.0` released! But the way `Connections` have been released does not conform to the spec and also regressed in functionality compared to `0.9`, by not having this test: https://github.com/graphql-python/graphene/blob/0.x/graphene/relay/tests/test_query.py#L69

Having this check: https://github.com/graphql-python/graphene/blob/master/graphene/relay/connection.py#L111-L114 in the `ConnectionField` makes it out of the box not work with `ObjectTypes`, which per spec are allowed to be used as `Connections`.  My PR https://github.com/graphql-python/graphene/pull/288 addresses this issue. This is important to us because with how things are right now we can not upgrade our apps and we would love to do this as soon as possible.

Also a connection field resolver still does not allow the returning of a `Promise`.

Thanks for all the great work done so far!

syrusakbary:
Hey Markus,

Thanks for opening the PR and this issue and sorry for replying so late.
I agree that the way connections are done now have a lot of things to improve, and actually I like most of the ideas that you have in the PR.

However before moving forward with it I think is worth to analyze how connections should work, trying to solve questions like:
* Where the connection resolution should live? Should we start transferring some resolver logic to the Connection type?
* How we should handle the "automatic" connection types creation? We should handle it at all? (as is right now)
* How we can handle different connections for same types? And then, how we handle name clashing?

Some of this questions are easy to solve if we are more explicit (as your PR points), but before choosing a definitive approach would be good to explore if there is any way that we could achieve simplicity and extensibility at the same time (as we did for `Node`s).

In general I think is worth to spend some time analyzing, discussing and getting a stable API for connections than releasing something that we might have to change after.

Please raise any ideas you have! (and thanks for your great work too!)

PS: Meanwhile I'm going to add support to the test you pointed to unblock your transition to `1.0` :)

Globegitter:
Hey Syrus,

Thank you for getting back with such a detailed answer. Yes, my PR is certainly not perfect, it was partially also me just getting to know that area of the code and if I would make another PR I would probably do it differently. 

And yes, I think it is very good that you are asking these questions and before anything is rewritten we should answer them.

 - Having some connection resolution on the type could make sense. Was there anything you had in mind already?
[...]

Issue #180, opened 2016-05-23
True cursor pagination for Django


mjtamlyn:
## Problem

The default pagination approach used for Django querysets executes a `.count()` on the query. For tables with hundreds of thousands or millions of rows, this is very slow on Postgres. It also assumes that the set is stable, which somewhat defeats the point of using a cursor based pagination system in the first place.
## Possible solution

I released today [`django-cursor-pagination`](https://github.com/photocrowd/django-cursor-pagination) which provides a paginator class which fulfils the [relay cursor connection spec](https://facebook.github.io/relay/graphql/connections.htm#EdgesToReturn%28%29). It is available on PyPI.

Below is a pretty simple approach to using this in Graphene, with inspiration taken from [graphene-gae](https://github.com/ekampf/graphene-gae/blob/master/graphene_gae/ndb/fields.py) - thanks to @ekampf. It would want some checks around `full_args`, and possibly a means to explicitly set the ordering. It would also need testing and documenting.

Should we use this in graphene? There are some caveats of using this approach. Most importantly, the ordering fields in my current implementation must uniquely determine the items in the queryset. ([Discussion of this restriction](https://github.com/photocrowd/django-cursor-pagination/issues/1)). Perhaps we should provide both options?

``` python
import graphene
from cursor_pagination import CursorPaginator

def connection_from_cursor_paginated(queryset, args, connection_type, edge_type, pageinfo_type, **kwargs):
    paginator = CursorPaginator(queryset, queryset.query.order_by)
    full_args = dict(args, **kwargs)
    page = paginator.page(**full_args)

    edges = []
    for item in page:
        edge = edge_type(node=item, cursor=paginator.cursor(item))
        edges.append(edge)

    return connection_type(
        edges=edges,
        page_info=pageinfo_type(
            start_cursor=paginator.cursor(page[0]),
            end_cursor=paginator.cursor(page[-1]),
            has_previous_page=page.has_previous,
            has_next_page=page.has_next,
        )
    )


class CursorPaginatedConnection(graphene.relay.types.Connection):
    @classmethod
    def from_list(cls, queryset, args, context, info):
        connection = connection_from_cursor_paginated(queryset, args, connection_type=cls, edge_type=cls.edge_type, pageinfo_type=graphene.relay.PageInfo)
        connection.set_connection_data(queryset)
        return connection


class CursorPaginatedConnectionField(graphene.relay.ConnectionField):
    def __init__(self, *args, **kwargs):
[...]

Issue #62, opened 2015-12-04
Dealing with cursors in ConnectionField resolve functions


bigblind:
In the Star Wars relay example, it looks like you expect the `resolve_ships` method of the `Faction` class to return every ship. While this is feasible with a small dataset like that, it's not an assumption you can make. Does the resolve function get access to the `first` and `after` argument? Also, how do we supply pagination info?

Looking at the code, it looks like `connection_from_list` deals with the resolved data as a complete list. Is there a way to get around this? If the resolve function only gets the arguments, but no way to supply pagination info, should we just return 1 more object than was requested, so that `connection_from_list` knows there's more and can correctly create the pagination info?

syrusakbary:
Hi @bigblind,
`connection_from_list` expect an iterable (this iterable had to implement `len` and a item getter like `iterable[20:30]` though). Could be a `list`, could be a lazy list, or anything that you could iter through.
However, we don't iter through the iterable until the end (when we show the results).

In the example of [Starwars relay](https://github.com/graphql-python/graphene/blob/master/examples/starwars_relay/schema.py) the `resolve_ships` return a complete list. But the [swapi example](https://github.com/graphql-python/swapi-graphene/blob/master/starwars/schema.py#L146) the method returns an iterable.

So, for example, when you are using django querysets, the list is only "resolved" (aka queried in the DB) when we iter though it (when we return the resolved result), and after the list is sliced.

So it will already do in the most optimal way :)

bigblind:
Can I just create my own field with a type of `Connection`? The database I'm using, rethinkdb, allows me to do much more optimized queries with cursors than with indices based on `len`. Basically, with the approach that you use, I need to know the index of every item in the database. With cursors, a rethinkdb query would look like this:

`r.db("my_db").table("users").between(cursor, r.maxval).limit(first+1).run(connection)``

If you're not interested in rethinkdb, I don't expect you to look up how this code works, but I wanted to give a use case for actually doing this.

The reason I'm fetching `first+1` items, is so that I know there's a next page of results. if the query returns `first` items or fewer, I know there are no more items.

syrusakbary:
Good catch. I will *rethink* better how we can handle this.
Probably moving some code from the `graphql_relay` Python library to `graphene` and wrapping in some class where people could manage it easily is the path to go!
Will keep this issue open until is resolved.

ekampf:
Similar case here working with AppEngine's NDB (would be the same for HBase or most other nosqls I guess).
[...]

PR #347, opened 2016-11-03
New Connection API

syrusakbary:
This PR improves the `relay.Connection` APIs in Graphene.
Is inspired on the work of @Globegitter in #288 and following #304.

Things that this PR should achieve:
* [x] Remove automatic `.Connection` creation in Nodes in pro of cleaner and more explicit code
* [ ] Remove the `graphql-relay` dependency.
* [ ] One connection type could have different resolving strategies (different cursors / In a fully list, in a iterable without an explicit length = django queryset, ...)

## Related issues
#304 #180 #62

### DIFF
--- a/graphene/relay/connection.py
+++ b/graphene/relay/connection.py
@@ -96,6 +96,10 @@ class Connection(six.with_metaclass(ConnectionMeta, ObjectType)):
         return ConnectionField(cls, *args, **kwargs)
 
     @classmethod
+    def is_type_of(cls, root, context, info):
+        return isinstance(root, cls)
+
+    @classmethod
     def connection_resolver(cls, resolved, args, context, info):
         assert isinstance(resolved, Iterable), (
             'Resolved value from the connection field have to be iterable or instance of {}. '
@@ -132,7 +136,7 @@ class ConnectionField(Field):
     def connection_resolver(cls, resolver, connection_type, root, args, context, info):
         resolved = resolver(root, args, context, info)
 
-        if isinstance(resolved, connection_type):
+        if connection_type.is_type_of(resolved, context, info):
             return resolved
 
         on_resolve = partial(connection_type.connection_resolver, args=args, context=context, info=info)
--- a/graphene/relay/connection.py
+++ b/graphene/relay/connection.py
@@ -90,17 +90,39 @@ class ConnectionMeta(ObjectTypeMeta):
 
 
 class Connection(six.with_metaclass(ConnectionMeta, ObjectType)):
-    pass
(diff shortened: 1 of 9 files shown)


graphene/relay/connection.py
```diff
@@ -90,65 +90,57 @@ class ConnectionBase(AbstractType):
...
     @classmethod
     def connection_resolver(cls, resolver, connection_type, root, args, context, info):
         resolved = resolver(root, args, context, info)
 
-        on_resolve = partial(cls.resolve_connection, connection_type, args)
+        if isinstance(resolved, connection_type):
+            return resolved
+
+        on_resolve = partial(connection_type.connection_resolver, args=args, context=context, info=info)
         if isinstance(resolved, Promise):
```
Globegitter: The only minor problem I have with this is that the instance check does not allow for duck-typing, which goes in line with https://github.com/graphql-python/graphql-core/pull/91 and is a common principle in python

graphene/relay/connection.py
```diff
@@ -90,65 +90,57 @@ class ConnectionBase(AbstractType):
...
+            *args,
+            **kwargs
+        )
+
     @classmethod
     def connection_resolver(cls, resolver, connection_type, root, args, context, info):
         resolved = resolver(root, args, context, info)
 
-        on_resolve = partial(cls.resolve_connection, connection_type, args)
+        if isinstance(resolved, connection_type):
```
Globegitter: Support for duck-typing would be nice here as well but I think it would probably be far less used and more a nit-pick from my side,

Globegitter:
Liking the direction this is heading so far :)

tsunammis:
Hello @syrusakbary 
Since your commit -> [760ccc8](https://github.com/graphql-python/graphene/commit/760ccc83587396df9a2cbf5806b7f4fa73ba620e), Is it not possible to use async resolver method for a `relay.ConnectionField`, isn't it ?

I get this error ->  `AssertionError: Resolved value from the connection field have to be iterable or instance of SectionConnection. Received "<coroutine object get_sections at 0x7f0de4b14518>"`

ajhyndman:
Hey guys.  I'm wondering what the status is on this PR.  It seems like some of the documented Connection APIs might be out of date?

I'm getting errors that look like this when I try to declare a property with type `ConnectionField(MyNode)`:

```
16, in is_node
    if not issubclass(objecttype, ObjectType):
TypeError: issubclass() arg 1 must be a class
```

syrusakbary:
Closing in favor of #500

PR #352, opened 2016-11-05
Add link to official GraphQL docs in Getting Started 

ttz21:
I am contributing to the Graphene docs for the first time.  

This commit simply adds a reference to the official GraphQL docs - I haven't been able to build it locally due to graphql-python/graphene-python.org#8  but as it is a simple change, it should be ok.


syrusakbary:
Makes sense!
Merging :)

PR #361, opened 2016-11-10
Make node arg options and default required to True for GlobalID.

Globegitter:
This is pulling out some bits of #282 to make it a focused PR just trying to fix one thing. See discussion in in #282. We are most of the time overriding `GlobalID` and so it just makes it far less clean having to write `GlobalID(node=Node, required=True)` all the time, also somewhat surprising, as it was an undocumented breaking change between `0.x` and `1.x`.

Let me know if there are still any issues with that :)

### DIFF
--- a/graphene/relay/node.py
+++ b/graphene/relay/node.py
@@ -35,9 +35,9 @@ def get_default_connection(cls):
 
 class GlobalID(Field):
 
-    def __init__(self, node, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


syrusakbary:
Tests covering what this PR improves would be great, so this things are not missed in future refactors! :)

Globegitter:
@syrusakbary ha - you where a tiny step ahead of me. Just pushed the tests :)

PR #363, opened 2016-11-12
Add OmniScalar type

jamesboehmer:
Using the OmniScalar type, you can define a field where the scalar type is nondeterministic.  For example, take this Elasticsearch term aggregation on the `_type` and `year` fields in the index:

```
{
  "_type": {
    "buckets": [
      {
        "key": "article",
        "doc_count": 367086
      },
      {
        "key": "video",
        "doc_count": 40046
      }
    ],
    "sum_other_doc_count": 0,
    "doc_count_error_upper_bound": 0
  },
  "year": {
    "buckets": [
      {
        "key": 2015,
        "doc_count": 2963
      },
      {
        "key": 2016,
        "doc_count": 2059
      }
    ],
    "sum_other_doc_count": 357862,
    "doc_count_error_upper_bound": 614
  }
}
```

The `key` field is either a `String` or an `Int` depending on how it was mapped in Elasticsearch.  We'd like to make the aggregations part of the GraphQL schema, so instead of requiring the developer to define a union of object types with varying scalars for the `key` field, we simplify the schema by declaring only `key = OmniScalar()`.

### DIFF
--- a/graphene/__init__.py
+++ b/graphene/__init__.py
@@ -60,6 +60,7 @@ if not __SETUP__:
         'Float',
         'Enum',
         'Boolean',
+        'OmniScalar',
         'List',
         'NonNull',
         'Argument',
--- a/graphene/__init__.py
+++ b/graphene/__init__.py
@@ -26,7 +26,7 @@ if not __SETUP__:
         InputField,
         Schema,
         Scalar,
-        String, ID, Int, Float, Boolean,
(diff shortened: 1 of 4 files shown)


syrusakbary:
Hi @jamesboehmer,

I think things like a `OmniScalar` could be useful for other developers, but I think is better if is as an independent package as solves a very specific use case and could mislead the developers on which one they should use -`String` or `OmniScalar`- as both would work for "String" fields.

PR #367, opened 2016-11-15
Allow the `Union` type to be used in relay connections

tangerilli:
The union node needs a `Connection` type for things to work, so I create one for each type of union in the union meta class.  Maybe there's a better way to handle it?


syrusakbary:
Hi @tangerilli,

The way Connections work will be changed by #347 (removing the need of an `ObjectType` for have a `Connection` on it).

You can still make `Connection` work with `Union`, doing something like:

```python
class MyUnion(Union):
    class Meta:
        types = (MyObjectType1, MyObjectType2)

class MyUnionConnection(Connection):
    class Meta:
        node = MyUnion

class Query(ObjectType):
    objects = ConnectionField(MyUnionConnection)

    def resolve_objects(self, args, context, info):
        return [MyObjectType1(), MyObjectType2()]
```

tangerilli:
@syrusakbary thanks, that worked!

PR #372, opened 2016-11-22
Added Time Type

pizzapanther:
Also changed one DateTime method to a class method so the parser doesn't have to be updated in two places if there is ever a change.


pizzapanther:
Need a test?

syrusakbary:
The PR looks good!
Adding tests for ensuring that everything works as expected is a good idea before merging :)

PR #379, opened 2016-11-30
(relay connection) promisify result from async method

tsunammis:
With the current state of Graphene, is it not possible to use an `async` method as a **resolver**.

**Reason**:

The resolver can not handle coroutine, it works only with expected data type and also [with "Promise" object](https://github.com/graphql-python/graphene/blob/master/graphene/relay/connection.py#L145).

**Error message**:

```
AssertionError: Resolved value from the connection field have to be iterable or instance of PeopleConnection. Received "<coroutine object About.resolve_people at 0x7faad9c93a40>"
```

**Example**:

So, if you try with this piece of code

```python
import graphene
from graphene import relay

class Person(graphene.ObjectType):
    name = graphene.String()


class PeopleConnection(graphene.Connection):
    class Meta:
        node = Person


class About(graphene.ObjectType):
    class Meta:
        interfaces = (relay.Node, )

    people = relay.ConnectionField(PeopleConnection)

    @classmethod
    def get_node(cls, *_):
        return {}

    async def resolve_people(self, args, ctx, *_):
        return [People(name="Steve")]
```

You will get this error:
```
AssertionError: Resolved value from the connection field have to be iterable or instance of PeopleConnection. Received "<coroutine object About.resolve_people at 0x7faad9c93a40>"
```

**Suggested method:**

Update `connection_resolver` in relay/connection module to promisify async methods.

### DIFF
--- a/graphene/relay/connection.py
+++ b/graphene/relay/connection.py
@@ -12,6 +12,7 @@ from ..types import (AbstractType, Boolean, Enum, Int, Interface, List, NonNull,
 from ..types.field import Field
 from ..types.objecttype import ObjectType, ObjectTypeMeta
 from ..types.options import Options
+from ..utils.compat_asyncio_promise import promisify
 from ..utils.is_base_type import is_base_type
 from ..utils.props import props
 from .node import is_node
@@ -139,7 +140,7 @@ class IterableConnectionField(Field):
 
     @classmethod
     def connection_resolver(cls, resolver, connection_type, root, args, context, info):
-        resolved = resolver(root, args, context, info)
+        resolved = promisify(resolver(root, args, context, info))
 
         on_resolve = partial(cls.resolve_connection, connection_type, args)
         if isinstance(resolved, Promise):
--- a/graphene/utils/compat_asyncio_promise.py
+++ b/graphene/utils/compat_asyncio_promise.py
@@ -1,0 +1,12 @@
+import promise
+
+def promisify(resolved):
+    try:
+        import asyncio
+        if isinstance(resolved, asyncio.Future) or \
+            asyncio.iscoroutine(resolved):
+            return promise.promisify(resolved)
+    except:
+        pass
+
+    return resolved


tsunammis:
I saw [this commit](https://github.com/graphql-python/graphene/pull/347/commits/e6733d56b20bc6c8a5c065b1bcc03b6cd2a1feb2), It should be good to merge it, to fix the issue, isn't it ?

tsunammis:
cc @syrusakbary

tsunammis:
I have hesitated to cherry-pick your commit [e6733d5](https://github.com/graphql-python/graphene/pull/347/commits/e6733d56b20bc6c8a5c065b1bcc03b6cd2a1feb2), but your method `is_thenable` doesn't handle an object type which is equals to `coroutine`.

So, I think that you should add `asyncio.iscoroutine(resolved)` in `is_thenable` as you handle this type in the `promisify` method.

@syrusakbary What do you think ?

syrusakbary:
Hi @tsunammis,

At first thank you for contributing!
As you commented this were the issues:
* `is_thenable` returned `False` for coroutines, even if the `promisify` method [was able to handle it correctly](https://github.com/syrusakbary/promise/blob/master/promise/promise.py#L441-L442).
* Relay Connections were not accepting Promise-like objects / coroutines.

I fixed each separately in their corresponding packages:
* Fixed `is_thenable` for accepting coroutines (inside the `promise` package): https://github.com/syrusakbary/promise/commit/8b0d5da3dca2d99671b820478ef74353ac24c97b
* Relay connections now promisify if the object is thenable https://github.com/graphql-python/graphene/commit/3df62d26a7962f1273df972da21b25ac75bc9cf3

I've released a new version of promise: [1.0.1](https://github.com/syrusakbary/promise/releases/tag/v1.0.1) and graphene: [1.1.3](https://github.com/graphql-python/graphene/releases/tag/v1.1.3) that have this changes.

Please let me know if everything works as expected with the new versions! :)

tsunammis:
Hello @syrusakbary 

Thanks for your time. It works fine now.

Next time, I will take some time to create PR on each projects ;-)

Have a nice day.

syrusakbary:
Great! Closing the PR :)

Issue #316, opened 2016-10-06
v1.0: Lazily load model class in Field and relay.ConnectionField as in v0.x


JimVanEeden:
Currently rewriting our GraphQL API to work with graphene 1.0 but I'm coming across a big problem:

In order to create the models, I've written a construction class that will convert the actual database models to models used by graphene. There arises a problem however when creating relations on the models using the Field and relay.ConnectionField classes. In graphene 0.x I would be able to provide the name string as argument to Field and relay.ConnectionField, but apparently this is not possible anymore in 1.0 (I'm wondering why you took this out as well). 

It's impossible to import the model classes in the construction class due to circular dependencies. I'm hoping there is another solution to this problem, other than completely rewriting everything to something that is not logically structured at all anymore. Is this something that is going to change in the future or is there a fix available?

jameswyse:
I've ran into a similar problem, I need to include the parent when using RANGE_ADD mutations but `graphene.Field('Viewer')` no longer works

syrusakbary:
Hi @JimVanEeden @jameswyse, in `v1.0` you can use lambdas in a Field.

A simple example:

```python
class Viewer(graphene.ObjectType):
    viewer = graphene.Field(lambda: Viewer)
```

JimVanEeden:
Yes I got that from the documentation, the problem is that I can't import the classes. The only solution is to put all the models and the construction class in one file, which is highly undesirable.

jameswyse:
Same problem here. I'm gonna try using `__import__` or `importlib.import_module` inside the lambda. Not ideal either but it should do the trick.

jameswyse:
@JimVanEeden It works :)

Here's some example code:

```python
import graphene
from importlib import import_module
from graphql_relay.connection.arrayconnection import offset_to_cursor
from something_app.models import Something
from something_app.schema import SomethingNode


class CreateSomething(graphene.relay.ClientIDMutation):
    class Input:
        name = graphene.String()
        
    viewer = graphene.Field(lambda: import_module('your_app.schema').Viewer)
    something_edge = graphene.Field(SomethingNode.Connection.Edge)
    success = graphene.Boolean()

    @classmethod
    def mutate_and_get_payload(cls, input, context, info):
        name = input.get('name')

        something = Something(name=name)
        something.save()

        # hacky but works for now
        cursor = offset_to_cursor(Something.objects.count() - 1)
        edge = SomethingNode.Connection.Edge(node=something, cursor=cursor)

[...]

PR #381, opened 2016-12-02
fix so partial functions can be passed to field, solves issue #316

JimVanEeden:


syrusakbary:
Tests reflecting this improvement would help to not repeat the issue in the future! :)

PR #410, opened 2017-01-27
Allow class methods to be used as a field source

docmatrix:
Static methods are currently working, but it can be useful to allow class methods to work also, particularly in the context of django-graphene and using DjangoObjectType where you may have properties or methods on fatter models.


syrusakbary:
Makes sense! Thanks for the PR :)

PR #412, opened 2017-02-06
Added documentation on how to accept files in a mutation

BossGrand:
Added documentation that describes how to create a mutation that takes in files

Also fixed a few documentation bugs, there was a code type graphql instead of json, and a warning for a header underline that was too short


syrusakbary:
Looks great! Thanks for contributing :)

PR #415, opened 2017-02-16
Update README.md

helfer:
Mention that Graphene works great with both Relay and Apollo. I also took the liberty of turning both of these into links.


syrusakbary:
Hey Jonas, I improved a little bit the PR, placing the Apollo mention in the integrations part!

PR #435, opened 2017-03-11
Add documentation about Graphene + Apollo

helfer:
Hi @syrusakbary, I've added a section about Apollo with useful links and instructions for how to set up batching to the docs. I think this will go a long way in ensuring that people don't get confused about whether Graphene works with Apollo or not 🙂

### DIFF
--- a/docs/apollo/batching.rst
+++ b/docs/apollo/batching.rst
@@ -20,7 +20,7 @@ Graphene-Django has built-in support for transport batching, which can be enable
     from graphene_django.views import GraphQLView
 
     urlpatterns = [
-        url(r'^graphql', GraphQLView.as_view(graphiql=True, batch=True)),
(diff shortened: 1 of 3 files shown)


helfer:
@syrusakbary Ping. It would be great if this could be merged before we release Apollo 1.0!

vincentfretin:
@helfer I don't use Django but I looked at the code in graphene_django/views.py
There is this line in the `__init__`:
```
assert not all((graphiql, batch)), 'Use either graphiql or batch processing'
```
It seems that you can't enable graphiql with batch enabled, right? So it seems your example is wrong.
If this is Django specific, it may be a better idea to include this doc in https://github.com/graphql-python/graphene-django

helfer:
Thanks @vincentfretin, I just fixed it!

helfer:
@syrusakbary *bump*

PR #449, opened 2017-04-10
First version of the Graphene test client and snapshots 💪

syrusakbary:
This PR adds a dummy test graphql `Client` that helps to test.

```python
from graphene.test import Client

def test_hey():
    client = Client(my_schema)
    executed = client.execute('''{ hey }''')
    assert executed == {
        'data': {
            'hey': 'hello!'
        }
    }
```

Also, uses [SnapshotTest](https://github.com/syrusakbary/snapshottest) for **testing the examples with less code**!

## How snapshot testing works?

First time executing the tests will create *automatically* the [snapshots directory](https://github.com/graphql-python/graphene/tree/features/test-client-and-snapshot-testing/examples/starwars/tests/snapshots) with the contents of the api client response.

Then, next time all snapshots will be compared with the previous value, raising an error if there is any mismatch.

This helps **incredibly** the amount of time required to write tests, as now the code required to make the tests is just:

```python
def test_hey(snapshot):
    client = Client(my_schema)
    snapshot.assert_match(client.execute('''{ hey }'''))
```

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -10,6 +10,7 @@ Contents:
    types/index
    execution/index
    relay/index
+   testing/index
 
 Integrations
 -----
--- a/docs/testing/index.rst
+++ b/docs/testing/index.rst
@@ -1,0 +1,90 @@
+===================
+Testing in Graphene
+===================
+
+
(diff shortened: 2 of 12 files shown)


BossGrand:
Awesome pr!

I suggest adding some docs on passing context into the execute method

PR #452, opened 2017-04-13
Update basic schema with arguments

### LINKED ISSUE #450: Input args and documentation
Hi,

I'm trying to understand how take in arguments for fields but could not find it in documentation.

Also, the playground here seems to be broken as well: http://graphene-python.org/playground/

khankuan:
Closes #450


syrusakbary:
Great addition!

PR #455, opened 2017-04-20
add 'invalid' key in test client response

stegben:
Lots of my old projects required this QQ.


stegben:
require update snapshot. create another branch

PR #492, opened 2017-06-24
Update to_camel_case to handle leading and double-underscores.

adamdonahue:
### CHANGE SUMMARY

* Update `to_camel_case` to more gracefully handle multiple-underscore separators and leading underscores.

### NOTES

The current behavior is somewhat unintuitive.  For example:

```python
In [2]: to_camel_case('__all__')
Out[2]: '_All__'
```
The arbitrary removal of a single underscore is awkward.  I argue it makes more intuitive sense, at least in the way snake case variables are typically used, to treat single underscores specifically, and leave multiple underscores untouched.  (In my implementation they also 'reset' the camel casing, but that's more debatable.)

I've modified the code to change a couple of things:

* Leading underscores are left in place.  (`_all` remains `_all` and `_all_x` becomes `_allX`).
* Multiple underscores are not treated as snake case separators, and are, instead, left in place. Further snake case variable components following them are handled as if they were separate variables themselves in terms of conversion.  (`my__var_name` becomes `my__varName` not `my__VarName` -- though this is easy to change.)

*I'm open to feedback on any of these assumptions.  I think the most important change is preserving the leading underscores in a sensible way.*

### DIFF
--- a/graphene/utils/str_converters.py
+++ b/graphene/utils/str_converters.py
@@ -5,7 +5,7 @@ def to_camel_case(snake_str):
     """
     Return a camel-cased version of a snake-cased string.
 
-    Leading underscores and multiple-underscores are kept
+    Leading underscores and multiple consecutive underscores are kept
     intact.
 
     :param snake_str: A snake-cased string.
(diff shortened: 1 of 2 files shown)


adamdonahue:
Is anyone going to bother looking at this, or should I just close it?

adamdonahue:
Actually, I'll just close.   Feel free to reopen if outside contributions ever become of interest, though.

PR #508, opened 2017-07-21
Add totalCount field in ConnectionBase class

nxexox:
Add field totalCount in ConnectionBase class.
According to the official documentation of Graphql, the first level of pagination should be the totalCount parameter, which in graphenee is not yet available. This field is very useful and will make life much easier. I ask to include it in the main turnip, or to correct it. It seems that I did a little differently than in the general form of this repository, but in which case I will finalize it when I receive feedback.

### DIFF
--- a/graphene/relay/connection.py
+++ b/graphene/relay/connection.py
@@ -82,7 +82,12 @@ class ConnectionMeta(ObjectTypeMeta):
 
         class ConnectionBase(AbstractType):
             page_info = Field(PageInfo, name='pageInfo', required=True)
+            total_count = Int(name='totalCount', required=True)
             edges = NonNull(List(edge))
+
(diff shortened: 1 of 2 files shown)


ekampf:
@nxexox where do you see this in the Relay spec?  it only defines two fields as required -
 `pageInfo` and `edges` - https://facebook.github.io/relay/graphql/connections.htm#sec-Connection-Types.Fields

Also when you think about it, it doesnt make sense to have `totalCount` as part of the Connection spec because what do you do when you paginate a long nosql table?
Say the FB newsfeed for example... the totalCount is meaningless.

syrusakbary:
Closing this PR as `totalCount` is not part of the relay spec as mentioned by @ekampf.

PR #509, opened 2017-07-26
Documentation editing. Class function wasn't indented beneath its class

scotmatson:
I believe this line was intended to be indented.


syrusakbary:
Hi @scotmatson, first thanks for contributing!

The code was intended the way it was written (no indentation), however I think it was not a good example.
I've improved it with this commit: https://github.com/graphql-python/graphene/commit/f3bdd7de6901521bec7bd22e32743bf3c87fdb91

Please let me know if is still not clear!

Closing the PR

PR #518, opened 2017-08-03
Fix bug with "private" attrs

phil303:
I believe this is a bug but I'm not actually sure of the desired intentions of underscore prefixed kwargs.

When you pass an underscore kwargs, this currently raises an AttributeError because of the first conditional. The fix is to swap these conditionals to allow for setting of underscore attributes.

The way I'm using these is to set a private variable to be used in a resolver to fetch a first class resource. I'm not sure if that's the intended use case or if there's a better way to do that.


phil303:
Closing! Based on the tests, I understand how this is supposed to be used.

PR #519, opened 2017-08-04
Allow custom Meta attributes in subclasses of Interface

earlgreyz:
Similar to ObjectType we want to subclass Interface and add more Meta attributes. The usecase is to create SQLAlchemyInterface which will map model fields to the interface (useful when the database represents structures with inheritance)

### DIFF
--- a/graphene/types/interface.py
+++ b/graphene/types/interface.py
@@ -16,12 +16,19 @@ class InterfaceMeta(AbstractTypeMeta):
         if not is_base_type(bases, InterfaceMeta):
             return type.__new__(cls, name, bases, attrs)
 
-        options = Options(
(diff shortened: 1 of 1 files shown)


syrusakbary:
This PR should be no longer necessary in Graphene 2.0.
Now you can do:

```python
class SuperInterfaceOptions(InterfaceOptions):
    extra_option = None

class SuperInterface(Interface):
    class Meta:
        abstract = True

    @classmethod
    def __init_subclass_with_meta__(cls, extra_option=None, **options):
        meta = SuperInterfaceOptions(cls)
        meta.extra_option = extra_option
        return Interface.__init_subclass_with_meta__(cls, _meta=meta, **options)
```

Issue #554, opened 2017-09-29
_private_fields are not hidden and are still accessible via query


mrstork:
I couldn't really find anything about private fields in the documentation but it looks like there is support for this in the code, and there was an attempt to add to this functionality in #286. Assuming underscore private fields are a feature you want to support there is a bug in the implementation.

Currently instead of hiding the field `_secret` from the resulting object, it will instead come through as `Secret` after being treated with camel case correction. This means the only way to get private fields at the moment is to use object properties as [this line suggests](https://github.com/graphql-python/graphene/blob/v1.4.1/graphene/types/objecttype.py#L110)

I think this snuck through because the test in #286 is a little misleading. If instead the test set `_private_state = Field(MyType)` the test would fail.

Not at all certain, but I imagine a patch would add a check to [yank_fields_from_attrs](https://github.com/graphql-python/graphene/blob/v1.4.1/graphene/types/utils.py#L52) making sure the fields included don't start with an underscore.

I am happy to try to submit a patch if this is a feature you would like to continue to support. I would definitely switch over from object properties if this gets patched.

syrusakbary:
As per my comment on #555:

If you want to make the fields private, you just need to not instantiate with `Field` (or `String`...).

Like:
```python
class Query(ObjectType):
    hello = Field(String)
    _private = None
```

PR #555, opened 2017-10-02
Addresses #554 - Making _private fields hidden in query

mrstork:
I had some time so I made a patch for #554.
Not certain this is the best approach, not even sure if you want to support this feature, but this PR can be used as a place to discuss a potentially better solution.
Thanks!


mrstork:
Error from the travis logs: `The program 'py.test' is currently not installed. To run 'py.test' please ask your administrator to install the package 'python-pytest'`

syrusakbary:
Hi @mrstork.

If you want to make the fields private, you just need to not instantiate with `Field` (or `String`...).

Like:
```python
class Query(ObjectType):
    hello = Field(String)
    _private = None
```

Issue #558, opened 2017-10-04
Default resolver does not work for Enums


ksonj:
If no explicit resolver is given, graphene nicely resolves fields by trying to access the attribute of the same name on the current node. This does not work for Enum's though, as illustrated by the following example: 

```
import graphene


class SomeEnum(graphene.Enum):
    Bar = 0
    Baz = 1


class Query(graphene.ObjectType):

    foo = graphene.Field(SomeEnum)
    biz = graphene.Field(graphene.String)


class SomeObject:
    def __init__(self):
        self.foo = SomeEnum.Bar
        self.biz = 'foobar'


if __name__ == '__main__':
    obj = SomeObject()
    schema = graphene.Schema(Query)
    query = '{foo,biz}'
    result = schema.execute(query, root_value=obj)
    print(result.data['foo'])
    print(result.data['biz'])
```

This should print

    Bar
    foobar

is does however print

    None
    foobar

I would have expected this to either return`enum.value` or raise an exception, instead of silently returning `None`.

The obvious solution is to add a custom resolver, of course:

    class Query(graphene.ObjectType):
        foo = graphene.Field(SomeEnum)
        biz = graphene.Field(graphene.String)
        
        def resolve_foo(root, *args, **kwargs):
            return root.foo.value

I'd think we should either remove the implicit default-resolver altogether, or take care that all graphql-types are handled properly.

lincolnq:
I also have this problem, and it's a big blocker for me. However, I don't like the suggested PR. I think it would be better to fix it at the GraphQLEnumType level (because this is really not an issue with the resolver -- it's an issue with the enum type itself not recognizing instances of enum as valid values), but I'm not sure how exactly is the best way to do it.

kenashcraft:
@lincolnq The workaround should be straightforward:

```
class Query(gql.ObjectType):
    currency2 = graphene.Field(Currency)  // NOTE: Previous code was just 'Currency()'

    def resolve_currence2(self, info):
        return Currency.USD.value // NOTE: Previous code was just 'Currency.USD'
```

alexisrolland:
@lincolnq it works.
Would you know by chance how to return the actual list of currencies rather than just Currency.USD.value?

lincolnq:
@alexisrolland Your question is off-topic for this issue, but the list should be part of your schema. Depending on the client library you are using, you may already have a matching type definition already auto-generated for you. If you don't, then you could make a server request to the introspection schema in order to get the type definition.

[...]

PR #559, opened 2017-10-04
Add isinstance check for Enum in default resolver

ksonj:
This should fix https://github.com/graphql-python/graphene/issues/558 (at least for Python >= 3.4), by checking if the automatically resolved attribute is an Enum.


ksonj:
Should be fixed in graphql-core with https://github.com/graphql-python/graphql-core/pull/140

PR #578, opened 2017-10-25
Pin versions of upstream requirements for 1.4.1

whalesalad:
**Warning:** Obviously this cannot be merged. 

Not even sure what the strategy would be here but old Graphene releases prior to 2.0 are now completely broken since these versions were not pinned.

Long story short, because `graphql-core` is new and this package looks for `>=` 1.1, it takes 2.0 and breaks. 

My branch is a working copy of 1.4.1. If you still rely on 1.4.1 in production and haven't been able to dedicate time to upgrade your code for 2.0, this will work.

    pip install https://github.com/whalesalad/graphene/archive/v1-4-1.zip

You can either upgrade to 2.0 or use this for the time being.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -82,9 +82,9 @@ setup(
 
     install_requires=[
         'six>=1.10.0',
-        'graphql-core>=1.1',
-        'graphql-relay>=0.4.5',
-        'promise>=2.0',
+        'graphql-core==1.1',
+        'graphql-relay==0.4.5',
+        'promise==2.0.2',
     ],
     tests_require=tests_require,
     extras_require={


yarian:
I tried adding the following to my `requirements.txt` and it seems to have accomplished the same outcome:

```
graphql-core ~= 1.1
graphql-relay~=0.4.5
graphene ~= 1.0
graphene_sqlalchemy ~= 1.0
```

Definitely annoying that this just broke with no warning. :\

warabak:
Should graphene release a 1.4.2 version with a gated version of `graphql-core` (`graphql-core>=1.1,<2.0`)?
This would allow a single version bump from 1.4.x instead of manually including graphene's transitive dependencies.

0xCMP:
To help those googling: `"Object is not a Promise like object"`

I only found this issue via my co-worker.

ekampf:
Adding `graphql-core ~= 1.1` to `requirements.txt` solves the problem but I do think there should be a new `graphene` 1.x release that fixed its `graphql-core` requirements...
@syrusakbary  ?

syrusakbary:
Yup, let me do a quick release of `1.4.2` so we fix this :)
Versions will be pinned on future releases of Graphene thanks to #581, so this way we assure this issue is not longer repeated.

Thanks all for raising the issue.

syrusakbary:
Just published a new version of Graphene `1.4.2` with the pinned requirements.
https://pypi.python.org/pypi/graphene/1.4.2

Closing the PR :)

whalesalad:
Thanks for doing that.

In the future I would suggest usin `==` instead of greater than and less than. While the chances are less likely due to the trajectory of those other projects, there is still room for this mistake to happen again. 

When the phrase “pinning a version” is used, it’s implied that the versions will be exact and not a range.

PR #596, opened 2017-11-03
Set all fields on input object types

nparrish:
Previously, this only set fields that were explicitly provided in the query.  Optional input fields that were _not_ provided in the query were left as Graphene field definition types (such as `List` or `ID`).

### DIFF
--- a/graphene/types/inputobjecttype.py
+++ b/graphene/types/inputobjecttype.py
@@ -1,10 +1,13 @@
 from collections import OrderedDict
 
 from .base import BaseOptions, BaseType
+from .field import Field
 from .inputfield import InputField
+from .objecttype import ObjectType
+from .scalars import Scalar
+from .structures import List, NonNull
 from .unmountedtype import UnmountedType
 from .utils import yank_fields_from_attrs
-
 
 # For static type checking with Mypy
 MYPY = False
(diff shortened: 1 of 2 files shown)


graphene/types/inputobjecttype.py
```diff
@@ -23,8 +23,8 @@ class Meta:
 
     def __init__(self, *args, **kwargs):
         dict.__init__(self, *args, **kwargs)
-        for key, value in self.items():
-            setattr(self, key, value)
+        for key in self._meta.fields.keys():
+            setattr(self, key, self.get(key, None))
```
japrogramer: Looks good to me, doesn't really make sense to return the field unless it is later used to validate data generated or retrieve from somewhere else other than user input.

nparrish: Right...and returning a combination of fields and values is a bit confusing.

nparrish:
@syrusakbary: any objections to this?

syrusakbary:
@nparrish it looks great!
Before merging it would be needed to have some tests that test against the issue that this PR resolves, so we make sure this issue is no repeated in the future :)

nparrish:
@syrusakbary This became a little more complicated than I initially thought.  Handling complex inputs (like lists of non-null objects, for example) required a bit more work.  I think I'm now handling all the possible cases, but it's possible I missed something.

PR #622, opened 2017-12-09
[Docs] Functional middleware example

jkimbo:
Added an example of using a function for middleware instead of a class.


syrusakbary:
Thanks @jkimbo

PR #638, opened 2018-01-05
Dict or attr default resolver

jkimbo:
I've created a `dict_or_attr_resolver` so that you can pass dicts or objects to object types.

I've also set it as the default resolver since I can't think of a downside but let me know if you don't think thats a good idea.

### DIFF
--- a/graphene/types/resolver.py
+++ b/graphene/types/resolver.py
@@ -7,9 +7,10 @@ def dict_resolver(attname, default_value, root, info, **args):
 
 
 def dict_or_attr_resolver(attname, default_value, root, info, **args):
+    resolver = attr_resolver
     if isinstance(root, dict):
-        return dict_resolver(attname, default_value, root, info, **args)
-    return attr_resolver(attname, default_value, root, info, **args)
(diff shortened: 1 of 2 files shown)


jkimbo:
Just found https://github.com/graphql-python/graphql-core/pull/103 which looks related. Not sure where the best place to solve this is? @syrusakbary @jhgg

graphene/types/resolver.py
```diff
@@ -6,7 +6,14 @@ def dict_resolver(attname, default_value, root, info, **args):
     return root.get(attname, default_value)
 
 
-default_resolver = attr_resolver
+def dict_or_attr_resolver(attname, default_value, root, info, **args):
+    resolver = attr_resolver
+    if isinstance(root, dict):
```
ekampf: `resolver = dict_resolver if isinstance(root, dict) else attr_resolver` is more Pythonic IMHO

ekampf:
This looks fine to me. 
I want to fix the build in the master branch before starting to approve and merge PRs.
I'll try to get `master` ready to accept new PRs ASAP

PR #648, opened 2018-01-19
Changed dead links to Relay documentation

mborus:
https://facebook.github.io/relay/docs/graphql-relay-specification.html is no longer available,
https://facebook.github.io/relay/docs/en/graphql-in-relay.html seems to be the current version


syrusakbary:
The link should direct to https://facebook.github.io/relay/docs/en/graphql-server-specification.html rather than `graphql-in-relay.html` (as the latter is for the React-Relay integration).

I will do it in a direct commit.
Thanks for pointing the issue :)

PR #650, opened 2018-01-21
Use input_class as arguments if dict

SN9NV:
For two of my GraphQL mutations, I want to have my create and update functions use the same input parameters, but update with an additional "id" field. I strongly believe in "don't repeat yourself", so I would like to create one dict or class that they can both use/inherit, and be able to add the "id" to the update function. If I decide to add new parameters, I want both mutations to be affected with minimal code updates.

I tried a few things, but none worked. After looking at the code, I saw that there was a simple fix that would allow me to use a dictionary and be able to do this:
```python
UserParameters = {
    "name": graphene.String(),
    "email": graphene.String(),
    ...
}

class UserCreate(graphene.Mutation):
    Arguments = {
        **UserParameters,
        "password": graphene.String()
    }
   ...

class UserUpdate(graphene.Mutation):
    Arguments = {
        **UserParameters,
        "id": graphene.ID(required=True)
    }
    ...
```

I don't know if there was another way to achieve this result with the current method of creating the argument inputs.
If anyone has a more Pythonic or elegant way to achieve this result, I'm open to ideas.

### DIFF
--- a/graphene/types/mutation.py
+++ b/graphene/types/mutation.py
@@ -54,7 +54,10 @@ class Mutation(ObjectType):
                     ).format(name=cls.__name__))
 
             if input_class:
-                arguments = props(input_class)
+                if isinstance(input_class, dict):
+                    arguments = input_class
+                else:
+                    arguments = props(input_class)
             else:
                 arguments = {}
 


jkida:
Still very new to graphql/graphene but i found this works.
```
user_input = {
    'name': graphene.String(required=True, description="Full Name"),
    'email': graphene.String(description="Email"),
    'username': graphene.String(required=True, description="Username")
}

class AddUser(graphene.Mutation):
    class Meta:
        arguments = user_input

    user = graphene.Field(User)

    @classmethod
    def mutate(self, *args, **user):
        user = UserModel(**user)
        session.add(user)
        
        return AddUser(user=user)
```

ProjectCheshire:
I've set up mixins for sets of fields that I use with my mutation /subscription argument classes, and even the object type. Then I include those with the class, and any additional fields required. It works well for the object type for resolvers that 'just work' without additional defs, eg 1:1 mapping of the field name from the root/model. If I need to define a custom resolver, then I set that field explicitly.

PR #653, opened 2018-01-24
Update middleware example to support introspection

dkleissa:
In the `timing_middleware` example, introspection queries will fail due to `Schema` and others not having `_meta` attributes. API will work and tests pass but introspection will fail, which can be quite confusing to the developer. Simple update makes sure the `root` variable has a `_meta` attribute before accessing it.


jkimbo:
Nice catch!

PR #673, opened 2018-02-17
Fix bug when setting a Relay ConnectionField to be required

jkimbo:
Came across this when setting up Relay Connections in my schema. Worked around it by subclassing ConnectionField but thought I'd contribute the fix upstream.

### DIFF
--- a/graphene/relay/connection.py
+++ b/graphene/relay/connection.py
@@ -115,9 +115,6 @@ class IterableConnectionField(Field):
 
     @classmethod
     def resolve_connection(cls, connection_type, args, resolved):
-        if isinstance(connection_type, NonNull):
-            connection_type = connection_type.of_type
-
         if isinstance(resolved, connection_type):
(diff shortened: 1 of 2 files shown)


graphene/relay/connection.py
```diff
@@ -108,10 +111,13 @@ def type(self):
         assert issubclass(connection_type, Connection), (
             '{} type have to be a subclass of Connection. Received "{}".'
         ).format(self.__class__.__name__, connection_type)
-        return connection_type
+        return type
 
     @classmethod
     def resolve_connection(cls, connection_type, args, resolved):
+        if isinstance(connection_type, NonNull):
```
syrusakbary: This will check the `connection_type` each time the field is resolved, however we can move this logic to `connection_resolver` so is just executed once :)

jkimbo: Done!

PR #679, opened 2018-02-27
Made DateTime types return GraphQLError on fail

anisjonischkeit:
This change makes it so that when an incorrectly formatted date string gets passed to a Date / Time argument a GraphQLError is returned rather than a GraphQLLocatedError. Since Date / Time are types, their errors should not be in the same class as errors in your application. This is also inline with how other types work in graphene (graphene.Int, graphene.Float)

In our application, we hide all GraphQLLocatedErrors from the client as we don't wish to leak application information. As such this change is crucial for the users of our application to get any feedback about what they did wrong.

### DIFF
--- a/graphene/types/datetime.py
+++ b/graphene/types/datetime.py
@@ -38,8 +38,10 @@ class Date(Scalar):
 
     @staticmethod
     def parse_value(value):
-        return iso8601.parse_date(value).date()
-
+        try:
+            return iso8601.parse_date(value).date()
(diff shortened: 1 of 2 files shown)


jkimbo:
@anisjonischkeit this looks good. Could you please add some tests first?

anisjonischkeit:
added tests

jkimbo:
@syrusakbary this looks ready to merge

Issue #684, opened 2018-03-14
Unused field in connections?


vaskinyy:
This is rather a question that an issue that we came up with when trying to implement pagination.

It seems that this field is not used anywhere in the code
https://github.com/graphql-python/graphene/blob/master/graphene/relay/connection.py#L132 ? Since it's the last field in the function we thought (at the first glimpse) that this is where actual connection data sits and we can modify it which led to no results. What's the actual use case of this field or can it be just removed so we don't modify the code that is not used?

jkimbo:
@vaskinyy yep I think you're right, `iterable` is not used on the connection object anymore. I've implemented a custom `ConnectionField` for django pagination at work and omitted the `iterable` argument and it worked fine. @syrusakbary correct me if I'm wrong.

jkimbo:
Also @vaskinyy a PR to remove it would be welcome

vaskinyy:
@jkimbo thanks for confirming! The PR is in place. 

We also tried to remove the field in our custom implementation and nothing happened.

PR #696, opened 2018-03-20
Remove unused field

vaskinyy:
Issue: #684 

`.iterable` is not used anywhere in the code so can be safely removed.


syrusakbary:
I believe `.iterable` is used by `graphene-sqlalchemy`

vaskinyy:
It seems that `graphene-sqlalchemy` has a variable called `iterable`, but it doesn't use this field https://github.com/graphql-python/graphene-sqlalchemy/search?utf8=%E2%9C%93&q=iterable&type=

PR #725, opened 2018-05-11
Added partial support to Dynamic type

nikordaris:


graphene/types/dynamic.py
```diff
@@ -11,7 +12,7 @@ class Dynamic(MountedType):
 
     def __init__(self, type, with_schema=False, _creation_counter=None):
         super(Dynamic, self).__init__(_creation_counter=_creation_counter)
-        assert inspect.isfunction(type)
+        assert inspect.isfunction(type) or isinstance(type, partial)
```
jkimbo: I think a better check would be to use `callable` then it would catch both cases.

nikordaris: You can't use callable because class types are callable.

nikordaris: I could do ```not inspect.isclass(type) and callable(type)```

nikordaris: Really though I was basing it off of https://github.com/graphql-python/graphene/blob/12d4dab7742c9408777380c390152d7cef0f838d/graphene/types/utils.py#L44

I wasn't sure if there were callable types we didn't want to support

jkimbo: Ah yeah you're right. What you've done then looks good. Just needs some tests.

nikordaris:
ok @jkimbo I've added a test for partial. let me know if there is anything else i should change

syrusakbary:
It goods look to merge for me @jkimbo, feel free to merge once you review it :)

jkimbo:
@syrusakbary I can't merge, I need write access to do that.

jkimbo:
This has been released in v2.1.1

PR #730, opened 2018-05-22
Add pre-commit tool to repository, and run on all files

dan98765:
Add pre-commit https://pre-commit.com/ and a config with a minimal set of hooks. Run on all files with `tox -e pre-commit` (also gets run as part of `tox`). I added a couple pep8 rules to the ignore list that were causing some not-pretty formatting and I didn't want to do any manual restyles as part of this PR. That said, I'm happy to change the config or ignored rules however the maintainers prefer. 

Similar PR for graphql-core: https://github.com/graphql-python/graphql-core/pull/178

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,17 @@
+-   repo: git://github.com/pre-commit/pre-commit-hooks
+    sha: v0.8.0
+    hooks:
+    -   id: autopep8-wrapper
+        args:
+        - -i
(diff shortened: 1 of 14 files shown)


syrusakbary:
Thanks for the PR @dan98765!
Merging :)

dan98765:
Thanks @syrusakbary! 
https://github.com/graphql-python/graphql-core/pull/178 is the exact same change but for graphql-core.

PR #734, opened 2018-05-25
Add reorder-python-imports pre-commit hook and run on all files

dan98765:


examples/starwars/tests/test_query.py
```diff
@@ -1,7 +1,6 @@
-from graphene.test import Client
-
 from ..data import setup
 from ..schema import schema
+from graphene.test import Client
```
jkimbo: Top level imports should be above relative imports

graphene/__init__.py
```diff
@@ -1,38 +1,40 @@
...
+from .types import Mutation
+from .types import NonNull
+from .types import ObjectType
+from .types import ResolveInfo
+from .types import Scalar
+from .types import Schema
+from .types import String
+from .types import Time
+from .types import Union
+from .types import UUID
```
jkimbo: I'm really not a fan of the imports being split up like this. How the imports were arranged is much clearer and easier to read.

syrusakbary:
I agree with @jkimbo. Replacing `reorder_python_imports` with `isort` (https://github.com/pre-commit/mirrors-isort ) should solve it

dan98765:
@jkimbo @syrusakbary thanks for the responses; I truly don't care which import sorter is used I just really like having one. :) 

I created https://github.com/graphql-python/graphene/pull/743 to do the same thing as this PR, but with  isort. I'll close this PR out.

PR #735, opened 2018-05-25
Add requirements-txt-fixer pre-commit hook and run on all files

dan98765:


dan98765:
@jkimbo you're right this isn't very useful yet. I'd like to try tossing the test reqs into a requirements-dev.txt file and doing some other changes to try and make it easier for first time contributors to get started; could we leave this open a bit until you've had a chance to consider those changes and then decide?

jkimbo:
Any reason why you prefer having the test requirements in a `requirements-dev.txt` file and not in the `setup.py` as they currently are?

PR #738, opened 2018-05-25
"type" is a builtin in python; switch some arguments called "type" to be "type_"

### LINKED ISSUE #700: Query cannot have a field named 'type'
https://github.com/graphql-python/graphene/blob/d46d8e8c33b1432ed664d94f432f357c44f3447e/graphene/types/field.py#L23

Because the first argument in the constructor is 'type', so a field named 'type' in 'extra_args' is not permitted:

TypeError: __init__() got multiple values for argument 'type'

Could by solved by renaming 'type' in the constructor to 'type_' (which is good python practice anyway for python builtins).

dan98765:
Closes #700

Also:
- Underscore prefix generally is understood to mean 'private' more than just 'not a builtin' so changed _type to type_ to be consistent. 
- Add `aniso8601` dep to tox.ini to get tests to pass

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -11,4 +11,3 @@ trim_trailing_whitespace = true
 [*.{py,rst,ini}]
 indent_style = space
 indent_size = 4
-
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,0 +1,2 @@
+[settings]
+known_third_party = aniso8601,graphql,graphql_relay,promise,pytest,pytz,pyutils,setuptools,six,snapshottest,sphinx_graphene_theme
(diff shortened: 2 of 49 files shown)


dan98765:
@jkimbo how do you feel about this branch?

morenoh149:
@dan98765 needs rebase

jkimbo:
Since it’s a breaking change it should probably be part of a 3.0 release (instead of 2.2 as I suggested earlier). Do you agree @graphql-python/governors ?

Cito:
@jkimbo I'm also using `type_` instead of `type` in graphql-core-next, so this would be consistent.

PR #741, opened 2018-05-26
Fix: Make `tox` stop failing

dan98765:
Tox stopped working due to recent changes; add in necessary dependencies to tox.ini so it passes again. 

With this branch, `tox` is green again.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -4,6 +4,7 @@ skipsdist = true
 
 [testenv]
 deps=
+    aniso8601
     pytest>=2.7.2
     graphql-core>=1.1
     promise>=2.0
@@ -11,10 +12,12 @@ deps=
     six
     blinker
     singledispatch
+    snapshottest
     mock
     pytz
     iso8601
     pytest-benchmark
+    pytest-mock
 setenv =
      PYTHONPATH = .:{envdir}
 commands=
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -4,17 +4,20 @@ skipsdist = true
...
     graphql-relay>=0.4.5
     six
     blinker
     singledispatch
+    snapshottest
     mock
     pytz
     iso8601
     pytest-benchmark
+    pytest-mock
```
jkimbo: Why do you have to define the dependencies here when they are already defined in the `setup.py`?

dan98765: This was the existing pattern in this repo; tox is pulling deps only from that list for running tests and not reading the list of test reqs from setup.py. 

I suggest accepting this PR so tox stops failing for now. Then I'd be happy to put out a followup PR to create a requirements-dev.txt file which can be the single source of truth for test requirements.

jkimbo: But why doesn't the tox file just have the `pip install -e ".[test]"` command that is outlined in the README? Then the whole `[testenv]` section just becomes:

```
[testenv]
commands = 
	pip install -e ".[test]"
	py.test --cov=graphene graphene examples
```

and there is no need to define requirements in multiple places.

dan98765: Sounds good; updated!

dan98765:
@jkimbo I think it would be good to update the contributing docs re: using tox. I don't think that should be a blocker on merging in this branch; I'll put out a separate PR to update the contributing docs.

jkimbo:
Yep totally agree. Just waiting for @syrusakbary to approve and then this PR can be merged.

PR #745, opened 2018-05-28
Create Makefile with several useful dev commands, and move dev requirements into requirements-dev.txt

dan98765:
1. Create Makefile with several useful dev commands, i.e. `make test` which runs `tox`, `make venv` which creates a development virtualenv called venv-graphene, `make clean` and `make install-hooks`. 
2. move dev requirements into `requirements-dev.txt` and delete test-related code from `setup.py`
3. Update .travis.yml to set up test environment by using the new `requirements-dev.txt` file

`make test` passes on this branch. The main thing I'm unsure of is A) any other fallout from removing the 'test' target in setup.py? B) if travis will complain at all about the .travis.yml change. 

The goal is for a new contributor to the repo to have easier start up process.
They clone repo, then run `make venv`, source that virtualenv, and then go to town writing code/tests while in a virtualenv which has benefits over installing things directly to their system.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -82,3 +82,6 @@ target/
 *.sqlite3
 .vscode
 .mypy_cache
+
+# Development virtualenv
+venv*
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ before_install:
 install:
 - |
   if [ "$TEST_TYPE" = build ]; then
-    pip install -e .[test]
+    pip install -r requirements-dev.txt
     python setup.py develop
   elif [ "$TEST_TYPE" = lint ]; then
     pip install flake8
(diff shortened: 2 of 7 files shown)


jkimbo:
@dan98765 you didn't answer my question as to why this structure is better than the current one? I'm not convinced this is clearer for new starters.

I'm fine with removing the `test` command in `setup.py` since I expect most people are familiar with just running pytest directly. I like the Makefile for helper steps but I don't think it should be creating a virtualenv for you. Outlining that you should probably be creating a virtualenv when developing should be part of the README.

Generally people should be able to get started in a couple of simple steps but I think that is already the case. So I'm happy to accept some small changes to the setup.py (like dropping the `test` option) and the introduction of the `Makefile` to help with some commands like installing pre-commit hooks, but I'm not going to approve a large change like this since I don't think it provides much benefit.

dan98765:
https://github.com/graphql-python/graphene/pull/765 to update README to encourage use of virtualenv

PR #748, opened 2018-05-30
Fix warning output

danpalmer:
Warning filtering is the responsibility of the application, not a library, and this current use causes all warnings from an application (at least those after this function is evaluated the first time) to print their contents.

This makes the library a better citizen in the Python ecosystem, and more closely matches what developers would expect.

(For what it's worth, we also can't start using this library without this patch because the logging is too verbose and may obscure more important warnings. We depend on being able to accurately control warning and logging output)


syrusakbary:
Makes sense, thanks for the PR :)

PR #764, opened 2018-06-09
Add requirements-dev.txt and remove test command in setup.py

dan98765:
Split out from https://github.com/graphql-python/graphene/pull/745

1. Adds a new requirements-dev.txt file to replace test requirements in setup.py
2. Remove `test` command in setup.py

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ before_install:
 install:
 - |
   if [ "$TEST_TYPE" = build ]; then
-    pip install -e .[test]
+    pip install -r requirements-dev.txt
     python setup.py develop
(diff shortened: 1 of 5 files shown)


syrusakbary:
@dan98765 having `test` in `extra_requirements` was intentional.
Some companies like to run graphene tests after install to test that graphene is working properly in their environment. Because of that, they require to be able to install requirements with `pip install graphene[test]`

dan98765:
@syrusakbary ah thanks for the context behind why `test` is there. 
Are you up for having `test` stay in `extra_requirements` and then have a `requirements-dev.txt` file in addition? Or would you prefer to drop this altogether?

Issue #50, opened 2015-11-25
ObjectType.resolve_type classmethod makes it impossible to have a field named 'type'


inklesspen:
The [`resolve_type`](https://github.com/graphql-python/graphene/blob/master/graphene/core/types/objecttype.py#L185) method on `ObjectType` conflicts with actually resolving a field called `type`. Given the importance of `resolve` method names, I think great care should be taken to make sure graphene's codebase does not use these method names unless necessary.

The below test fails with `GraphQLError("'Location' object has no attribute 'T'",)`; renaming the `resolve_type` method to `_resolve_type` solves the problem.

``` python
import graphene
from graphql.core.type import GraphQLEnumValue


LOCATION_TYPES = {
    'HQ': "Headquarters",
    'BRANCH': "Branch Location",
}


LOCATIONS = {
    '1': {
        'type': 'HQ',
        'address': '123 Main St',
        'city': 'Metropolis',
    },
    '2': {
        'type': 'BRANCH',
        'address': '55 Side St',
        'city': 'Smallville',
    },
}


LocationType = graphene.Enum(
    'LocationType',
    {k: GraphQLEnumValue(value=k, description=v) for k, v in LOCATION_TYPES.items()}
)


class Location(graphene.ObjectType):
    type = graphene.NonNull(LocationType)
    address = graphene.NonNull(graphene.String())
    city = graphene.NonNull(graphene.String())


class Query(graphene.ObjectType):
    location = graphene.Field(Location, id=graphene.ID(description='location id'))

    def resolve_location(self, args, info):
        loc_id = args.get('id')
        # error checking ommitted for brevity
        return Location(**LOCATIONS[loc_id])


schema = graphene.Schema(name="Sample", query=Query)

EXPECTED_SCHEMA = """
type Location {
  type: LocationType!
  address: String!
  city: String!
}

enum LocationType {
  BRANCH
  HQ
}

type Query {
  location(id: ID): Location
}
""".lstrip()

QUERY = """
fragment locFields on Location {
  type
  address
  city
}

query SomeQuery {
  hq: location(id: 1) {
    ...locFields
  }
  branch: location(id: 2) {
    ...locFields
  }

}
""".lstrip()

EXPECTED_RESPONSE = {
    'hq': {
        'type': 'HQ',
        'address': '123 Main St',
        'city': 'Metropolis'
    },
    'branch': {
        'type': 'BRANCH',
        'address': '55 Side St',
        'city': 'Smallville'
    },
}


[...]

PR #775, opened 2018-06-21
Rename `resolve_type` method

jkimbo:
Fixes https://github.com/graphql-python/graphene/issues/50

### DIFF
--- a/docs/types/interfaces.rst
+++ b/docs/types/interfaces.rst
@@ -154,7 +154,7 @@ Interfaces you might come across this error:
 
 This happens because Graphene doesn't have enough information to convert the
 data object into a Graphene type needed to resolve the ``Interface``. To solve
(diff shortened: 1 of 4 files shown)


syrusakbary:
I like this, however, let's make sure we write this down as a breaking change on the release, so people are aware the name has to be changed from `resolve_type` to `_resolve_type`.
(It will also require a minor version bump `2.2.0`)

honi:
Any update on this?

jkimbo:
For completeness the decision was taken to not make this change for the following reasons:

1. the workaround for it is quite easy:
```python
class MyInterface(Interface):
    type_ = String(name="type")
    def resolve_type_(_, info):
        return "foo"
```

2. `_resolve_type` is a bit ugly

3. I think we should be moving towards a @field decorator syntax for fields + resolvers. I’ve been using it my latest project and I quite like it.

Issue #783, opened 2018-06-30
Deduplication


femesq:
I was thinking in ways we could make [this #782](https://github.com/graphql-python/graphene/issues/782) work . I've also poked @syrusakbary on [twitter](https://twitter.com/syrusakbary/status/965660876333182976)...

The propose I was thinking about would improve the `middleware` class, adding a function just like Django's [new-style middleware](https://docs.djangoproject.com/en/2.0/topics/http/middleware/#upgrading-pre-django-1-10-style-middleware).

We would have something like this:
```python
class AuthorizationMiddleware(object):
    def resolve(self, next_mw, root, info, **args):
        ## resolver (request) processing here

    def process_response(self, next_mw, response, **args):
        ## result (response) processing here
```

Does it make any sense?

jkimbo:
@femesq you and @keeth made me curious about how you could add deduplication to graphene so I reimplemented graphql-deduplicator in python: https://github.com/graphql-python/graphene/pull/786

It's a pretty simple algorithm and it should work with the corresponding JS inflate method. It doesn't seem to help much with any of our queries at work but hopefully it's useful for someone.

jkimbo:
Closing this since the deduplicator functionality is now available in 2.1.3

Issue #782, opened 2018-06-28
Response deduplication


keeth:
Does graphene have an equivalent to [graphql-crunch](https://github.com/banterfm/graphql-crunch) or [graphql-deduplicator](https://github.com/gajus/graphql-deduplicator)?

If I wanted to write a wrapper that emitted a deduplicated response from graphene, could anyone suggest a starting point?  Seems like a straightforward way to save a lot of bandwidth.

Thanks!

jkimbo:
@keeth as far as I know there isn't a solution for response deduplication in the graphene ecosystem. I would love for someone to build one though!

It looks like you should be able to build something standalone that just takes the response from `execute`. 

E.g.:
```python
result = schema.execute('...')
deduped_data = deflate(result.data)
```

keeth:
Thanks! I'll close this issue for now.

jkimbo:
@keeth I've implemented both graphql-crunch (https://github.com/graphql-python/graphene/pull/787) and graphql-deduplicator (https://github.com/graphql-python/graphene/pull/786) in python. I didn't see any benefit on the queries we use at work but try them out and let me know if you see any benefits.

PR #786, opened 2018-07-01
Deduplicator

jkimbo:
Inspired by https://github.com/graphql-python/graphene/issues/783 and https://github.com/graphql-python/graphene/issues/782 I've implemented https://github.com/gajus/graphql-deduplicator in python.

### Usage:

```python
# You'll want to make sure that the query is not an introspection 
# query and that the client definitely wants the data to be deduplicated

result = schema.execute(query)
result.data = deflate(result)

# json encode result and serve to the client
```

### DIFF
--- a/graphene/utils/deduplicator.py
+++ b/graphene/utils/deduplicator.py
@@ -1,0 +1,51 @@
+from collections import defaultdict, Mapping, OrderedDict
+
+
+def nested_dict():
+    return defaultdict(nested_dict)
+
+
+def deflate(node, index=None, path=None):
+    if index is None:
+        index = {}
+    if path is None:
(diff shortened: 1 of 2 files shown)


gu3sss:
@jkimbo I am getting the below error in my code when I try the above recommendation,
`argument of type 'ExecutionResult' is not iterable"`
Here's my code,
`exec_result = schema.execute(str(query))`
        ```resp_body = {
                 "data": dict(deflate(exec_result)) if exec_result.data is not None else None,
       }```

I don't get the error if I use `deflate(exec_result.data)` instead. Nonetheless, the result isn't deduplicated. Any help is highly appreciated.

PR #794, opened 2018-07-08
Raising AssertionErrors instead of assert statements

Kacppian:
Code to use 

```
if something is not None:
    raise AssertionError("Something was unexpectedly None")
```

instead of 

`assert something is not None, "Something was unexpectedly None"`

### DIFF
--- a/graphene/relay/connection.py
+++ b/graphene/relay/connection.py
@@ -9,6 +9,7 @@ from ..types import Boolean, Enum, Int, Interface, List, NonNull, Scalar, String
 from ..types.field import Field
 from ..types.objecttype import ObjectType, ObjectTypeOptions
 from .node import is_node
+from ..utils.comparison_helper import raise_assertion_if
 
 
 class PageInfo(ObjectType):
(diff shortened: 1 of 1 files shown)


graphene/pyutils/version.py
```diff
@@ -44,8 +44,10 @@ def get_complete_version(version=None):
     if version is None:
         from graphene import VERSION as version
     else:
-        assert len(version) == 5
```
dan98765: I suggest only using `is` for comparing versus None, or identity comparisons. `==` is for comparing logical equivalence, which is going to be pretty much all numeric comparisons.

Some of the answers on https://stackoverflow.com/questions/2239737/is-it-better-to-use-is-or-for-number-comparison-in-python discuss it a bit more (1st SO page I googled).

graphene/relay/connection.py
```diff
@@ -107,20 +112,28 @@ def type(self):
...
     @classmethod
     def resolve_connection(cls, connection_type, args, resolved):
         if isinstance(resolved, connection_type):
             return resolved
 
-        assert isinstance(resolved, Iterable), (
-            "Resolved value from the connection field have to be iterable or instance of {}. "
-            'Received "{}"'
-        ).format(connection_type, resolved)
+        if not isinstance(resolved, Iterable):
```
dan98765: Rather than repeating the construction:
```
if BOOLEAN_EXPRESSION:
    raise AssertionError("...")
```
Could you DRY things up by writing a helper function that does this and use it across the code as a straight replacement for the `assert` statements?

Kacppian: I'll do that!

Kacppian: I am trying to do something like this -

Kacppian:
Closed. This was resolved in #808

PR #820, opened 2018-08-29
Update middleware.rst

dani0805:
I think a description of where the Middleware needs to go to be executed every time would be useful. It would have saved me some time.

### DIFF
--- a/docs/execution/middleware.rst
+++ b/docs/execution/middleware.rst
@@ -41,6 +41,18 @@ And then execute it with:
 
     result = schema.execute('THE QUERY', middleware=[AuthorizationMiddleware()])
 
+Or set it up in the default middleware to be executed for every query in settings.py:
+
+.. code:: python
+
+    GRAPHENE = {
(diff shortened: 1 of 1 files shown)


syrusakbary:
Hi @dani0805, this PR belongs to the [Graphene-Django docs](https://github.com/graphql-python/graphene), not to the normal Graphene Docs.
Please reopen there the PR for re-review :)

dani0805:
In graphene-django this part of the documentation is completely missing. Someone could extrapolate by looking how the debug middleware is deployed and then see the base documentation for general guide about how to develop a new middleware. 

But if you are not the one that installed graphene-django and you never saw it in settings, you might be trying to use the direct call example that is in the base documentation.

Would it make sense to:

1. repeat the explanation in graphene-django including deployment or,
2. reference the debug middleware django (and possibly flask if it even makes sense ) in the base documentation so that people using Django can find it easily

Let me know what you think and I can adapt it.

PR #821, opened 2018-08-29
Fix grammar in error message

danpalmer:


syrusakbary:
@danpalmer looks good. For tests to pass you will need to change also the following assertion:
https://github.com/graphql-python/graphene/blob/master/graphene/relay/tests/test_connection.py#L141

I will merge after! :)

syrusakbary:
👍 Merging

PR #832, opened 2018-09-09
Add basic type documentation for Relay fields

danpalmer:
This adds documentation in the API for `PageInfo`s and `Edges`.

This is useful to include in Graphene because `PageInfo` is always the same, and Edges always have the same format, so documentation for both can be created automatically.

### DIFF
--- a/graphene/relay/connection.py
+++ b/graphene/relay/connection.py
@@ -88,8 +88,22 @@ class Connection(ObjectType):
         _meta.node = node
         _meta.fields = OrderedDict(
             [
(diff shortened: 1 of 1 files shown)


danpalmer:
@syrusakbary I'm open to improving the documentation here, but I think this is a reasonable start.

We're trying to fully document our API, hence the PRs relating to enabling documentation for APIs built with Graphene.

syrusakbary:
Super useful PR towards improving documentation!
Thanks for the work @danpalmer, merging

PR #851, opened 2018-10-22
Improve dataloader doc

etandel:
I've recently encountered a bug caused by a dataloader returning the objects in a different order than the keys, so I thought it would be nice to have that explicitly said in the documentation.

### DIFF
--- a/docs/execution/dataloader.rst
+++ b/docs/execution/dataloader.rst
@@ -25,14 +25,13 @@ Create loaders by providing a batch loading function.
             return Promise.resolve([get_user(id=key) for key in keys])
 
 
-A batch loading function accepts an list of keys, and returns a ``Promise``
-which resolves to an list of ``values``.
+A batch loading function accepts a list of keys, and returns a ``Promise``
+which resolves to a list of ``values``.
 
 Then load individual values from the loader. ``DataLoader`` will coalesce all
 individual loads which occur within a single frame of execution (executed once
 the wrapping promise is resolved) and then call your batch function with all
 requested keys.
-
 
 
 .. code:: python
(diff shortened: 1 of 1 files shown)


docs/execution/dataloader.rst
```diff
@@ -47,6 +46,20 @@ requested keys.
...
+make sure that you then order the query result for the results to match the keys:
+
+
+.. code:: python
+
+   class UserLoader(DataLoader):
+       def batch_load_fn(self, keys):
+           positions = {k: pos for pos, k in enumerate(keys)}
+           users = User.objects.filter(id__in=keys)
+           return Promise.resolve(sorted(users, key=lambda p: positions[p.id]))
```
jkimbo: This won't work if you request a key that doesn't exist. The pattern we've been following is something like this:

```python
users = {user.id: user for user in User.objects.filter(id__in=keys)}
return Promise.resolve([users.get(id, None) for id in keys])
```

docs/execution/dataloader.rst
```diff
@@ -25,7 +25,7 @@ Create loaders by providing a batch loading function.
             return Promise.resolve([get_user(id=key) for key in keys])
 
 
-A batch loading function accepts an list of keys, and returns a ``Promise``
+A batch loading function accepts a list of keys, and returns a ``Promise``
 which resolves to an list of ``values``.
```
patrick91:
```suggestion
 which resolves to a list of ``values``.
```

etandel:
Thanks for the feedback @patrick91 @jkimbo!

I just implemented your suggestions, and rebased the branch.

syrusakbary:
Looks good to me. Feel free to merge once you re-review it @jkimbo @patrick91

Issue #699, opened 2018-03-27
Help with constructing an arbitrary ObjectType


jchang10:
Hi, I am still just learning. Looking for advice on an issue.

The simple question is, what is the best way to instantiate an ObjectType?

Here is a longer drawn out explanation for a simple question. I currently have something like below that returns a UserObjectType. Everything is great and user field resolves correctly. Returns the UserObjectType that is initialized by 'some_init_dictionary' which graphql takes care of on backend somewhere.

```
class NotesType(ObjectType):
    user = graphene.Field(UserObjectType)

    def resolve_user(self, args, context, info):
        return some_init_dictionary
```

Looking at examples, I decided I want to get fancy with Interfaces. Basically, I am trying to copy the example schema for StarWars and define the user field as a UserInterface. Then return an instance of UserObjectType.

```
class NotesType(ObjectType):
    user = graphene.Field(UserInterface)

    def resolve_user(self, args, context, info):
        return UserObjectType(**some_init_dictionary)

```
I simply replaced UserObjectType with UserInterface. That all works fine, too.

However, if **some_init_dictionary** has an extra field, then I get the "is an invalid keyword argument" error. In the original with UserObjectType, the extra field is ignored and everything merrily works fine.

Thus, am I missing something here? I would rather not have to do something like:
```
return UserObjectType(arg1='arg1', arg2='arg2', ... etc.)
```
and explicity call the constructor with each argument and field. I just want to duplicate how graphql handles the instantiation of UserObjectType as in the original version.

I hope that made sense. This should be a fairly simple answer/question. Thank you!

Jae

jchang10:
To answer my own question, I had to make sure to define a classmethod on UserObjectType named is_type_of(cls, value,context,info) to return a bool value on whether to accept the "returned data type" that your custom code returns from the resolver for your ObjectType.

See https://github.com/graphql-python/graphene/blob/d46d8e8c33b1432ed664d94f432f357c44f3447e/graphene/types/objecttype.py#L63

It's too bad there isn't better documentation. We have to rely on the code. However, the code isnt even really commented much, like the line above.

jkimbo:
@jchang10 another approach to resolving interface types is to define a (undocumented) classmethod `resolve_type(cls, instance, info)` method on the interface. This allows you to resolve the concrete type given some data. So in your example:

```python
class UserInterface(graphene.Interface):
	@classmethod
	def resolve_type(cls, data, info):
		if data['type'] == 'human':
			return Human
[...]

Issue #654, opened 2018-01-27
Document resolvers


Lucretiel:
I just learned, completely by accident, one of the most critically useful features of Graphene: resolvers can return anything, even for object resolvers! They don't have to return instances of graphene.ObjectType. So, this works:

```
class TestInstance:
    x = 1
    y = 2

class Test(graphene.ObjectType):
    x = graphene.Int()
    y = graphene.Int()

class Query(graphene.ObjectType):
    test = Test()
    def resolve_test(self, info):
        return TestInstance()
```

While this is hinted at in the ObjectType reference docs, there are no code examples of the pattern, nor documentation of *any* kind about the behavior of resolver functions. Documentation about this feature, and guidance about nested Objects, should be added.

I'm trying to draft something but I don't have a lot of time these days so I wanted to file an issue just in case

altaurog:
This doesn’t seem to be supported with use of schema interfaces.  There may be a way to do it, but a direct adaptation of the above doesn’t work (note that `test = Test()` is incorrect).

Take the following schema:
```python
import graphene

def resolve_test(*_):
    return TestInstance()

class TestInstance:
    id = 1
    x = 1
    y = 2

class Test(graphene.ObjectType):
    x = graphene.Int()
    y = graphene.Int()

    class Meta:
        interfaces = (graphene.relay.Node,)

    @classmethod
    def get_node(*_, **__):
        return resolve_test()

class Query(graphene.ObjectType):
    node = graphene.relay.Node.Field()
    test = graphene.Field(Test, resolver=resolve_test)

schema = graphene.Schema(query=Query)
```

Now we can execute the following query:
```graphql
{ test { id x y } }
```

And get this result:
```json
{
   "data": {
      "test": {
         "id": "VGVzdDox",
         "x": 1,
         "y": 2
      }
   }
}
```

But if you try:
```graphql
{ node(id: "VGVzdDox") { ... on Test { id x y } } }
```

You will get the following error:
`graphql.error.base.GraphQLError: Abstract type Node must resolve to an Object type at runtime for field Query.node with value "<test.graphql.testschema.TestInstance object at 0x7fa84c5b05c0>", received "None".`

Presumably the limitations placed on the resolved object are to ensure that the executor can determine its type, since it does not know the type a priori in this case.

altaurog:
> one of the most critically useful features of Graphene

In any event, I am curious what you are doing that makes this so useful.

jkimbo:
[...]

PR #868, opened 2018-11-26
resolve_type: Tests for Interface

tony:
Hi,

I am trying to get `def resolve_type` working with Interface.  My goal is to get this functioning (if it's the case it's not working). Forgive me if I'm misunderstanding how resolve_type works! Hopefully this test helps assure the the example we document in the interfaces.rst documentation will work.

As a way to create a minimal example, I've created a PR, but not sure if I'm correct on it yet. I think we should probably put it in test_query and also test for the resolved data + field. I also tried putting it in test_interface.py, but that may be too low level for resolve_type to run.

Here are some of the places I've researched prior to coming here:
- https://medium.com/the-graphqlhub/graphql-tour-interfaces-and-unions-7dd5be35de0d `resolveType` in graphql
- https://github.com/graphql-python/graphene/issues/699: Example of `resolve_type`
- https://github.com/rmosolgo/graphql-ruby/issues/1267 (resolve_type vs is_type_of: Note, we currently do have a test for is_type_of, but not resolve_type).
- Other related resolve_types mentions buried in the tracker: https://github.com/graphql-python/graphene/issues/654, https://github.com/graphql-python/graphene/pull/775

### DIFF
--- a/graphene/types/tests/test_interface.py
+++ b/graphene/types/tests/test_interface.py
@@ -1,5 +1,7 @@
 from ..field import Field
 from ..interface import Interface
+from ..objecttype import ObjectType
+from ..typemap import TypeMap, resolve_type
 from ..unmountedtype import UnmountedType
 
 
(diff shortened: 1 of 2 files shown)


tony:
This issue was due to me using a field named `type`. I'm abandoning this PR since my case is probably solved by #775

Issue #877, opened 2018-12-14
General wrongness working with Enums


avivey:
This is for graphene 2.1.3.

There are a bunch of issues working with enums, here are two that bite me right now:

1. When specifying `default_value`, we need to explicitly add `.name` or `.value`, else introspection breaks (see commented out code).
  - If specifying `.value`, the schema generated is wrong (See test 1 below).
  - If specifying `.name`, the schema is correct, but the argument passed to the function is a string that doesn't match any value (See test 1 and 2).
2. When working with python-backed enums, the value passed to the resolver doesn't match any sensible value (See test 3, below).

Test code: 
```python3
#! /usr/bin/env python
import graphene
import json
import enum

class Episode(graphene.Enum):
    NEWHOPE = 4
    EMPIRE = 5
    JEDI = 6

class BackendLengthUnit(enum.Enum):
    METER = 'm'
    FOOT = 'f'

LengthUnit = graphene.Enum.from_enum(BackendLengthUnit)

class Query(graphene.ObjectType):
    # desc = graphene.String(
    #    v=graphene.Argument(Episode, default_value=Episode.NEWHOPE),
    #    description='See ./broken-schema.py')
    desc1 = graphene.String(
        v=graphene.Argument(Episode, default_value=Episode.NEWHOPE.value),
        description='default value in schema is `4`, which is not valid. Also, awkward to write.')
    desc2 = graphene.String(
        v=graphene.Argument(Episode, default_value=Episode.NEWHOPE.name),
        description='Default value is correct in schema doc. awkward to write.')
    pyenum = graphene.String(
        v=graphene.Argument(LengthUnit),
        description='More fun with python-enum backed enum.')

    def resolve_desc1(self, info, v):
        return f'argument: {v!r}'

    def resolve_desc2(self, info, v):
        return f'argument: {v!r}'

    def resolve_pyenum(self, info, v):
        return (
            f'Argument: {v!r}. '
            f'Is GrapheneEnum foot: {v == LengthUnit.FOOT}. '
            f'Is Backend enum foot: {v == BackendLengthUnit.FOOT}. '
        )

schema = graphene.Schema(query=Query)
print('*** Schema:')
print(schema)

def qp(query):
    print(json.dumps(schema.execute(query).data, indent=2))

def qpf(query):
    print(json.dumps(schema.execute(query).to_dict(), indent=2))

print('*** Test 1: use default values:')

qp('{desc1, desc2}')
# Should be:
# {
[...]

Issue #756, opened 2018-06-05
assertion error in ast_from_value during introspection query when providing enum member as default value


DrPyser:
So I have an InputObjectType with an Enum field, and I want to provide a default value. The enum field is created from a pre-existing python enum:

```
MyGrapheneEnumType = graphene.Enum.from_enum(
    MyEnumType
)
class MyInput(graphene.InputObjectType):
    my_enum_field = graphene.Field(MyGrapheneEnumType, default_value=MyGrapheneEnumType.MEMBER)
```
I'm not sure it's the right way to do it(documentation would help), but I don't think an assertion error is the expected result.

If I then do an introspection query(I'm using insomnia which does them automatically), like this one:
```
query IntrospectionQuery {
      __schema {
        queryType { name }
        mutationType { name }
        subscriptionType { name }
        types {
          ...FullType
        }
        directives {
          name
          description
          locations
          args {
            ...InputValue
          }
        }
      }
    }

    fragment FullType on __Type {
      kind
      name
      description
      fields(includeDeprecated: true) {
        name
        description
        args {
          ...InputValue
        }
        type {
          ...TypeRef
        }
        isDeprecated
        deprecationReason
      }
      inputFields {
        ...InputValue
      }
      interfaces {
        ...TypeRef
      }
      enumValues(includeDeprecated: true) {
        name
        description
        isDeprecated
        deprecationReason
      }
      possibleTypes {
        ...TypeRef
      }
    }

    fragment InputValue on __InputValue {
      name
      description
      type { ...TypeRef }
      defaultValue
    }

    fragment TypeRef on __Type {
      kind
      name
      ofType {
        kind
        name
        ofType {
          kind
          name
          ofType {
            kind
            name
            ofType {
              kind
              name
              ofType {
                kind
                name
                ofType {
                  kind
                  name
                  ofType {
                    kind
                    name
                  }
                }
              }
            }
          }
        }
      }
    }
```
I get the following stack trace:
```
2018-06-05 03:05:09,963 - graphql.execution.executor - ERROR - An error occurred while resolving field __InputValue.defaultValue
Traceback (most recent call last):
  File "/home/drpyser/Documents/python/kioto_rest/.env/lib/python3.6/site-packages/graphql/execution/executor.py", line 311, in resolve_or_error
    return executor.execute(resolve_fn, source, info, **args)
[...]

PR #446, opened 2017-04-08
Improved enums comparison and getters

syrusakbary:
This PR adds the ability to compare `graphene.Enum` members with their values:

```python
class RGB(Enum):
    RED = 1
    GREEN = 2
    BLUE = 3

assert RGB.RED == 1
assert RGB.GREEN == 2
assert RGB.BLUE == 3
```

Also, adds a similar syntax to the native Python `Enum` getters:

```python
class RGB(Enum):
    RED = 1
    GREEN = 2
    BLUE = 3

assert RGB.get(1) == RGB.RED
assert RGB['RED'] == RGB.RED
```

@dewiniaid

Issue #444, opened 2017-04-01
Argument of type `SomeEnum` does not map back to `SomeEnum.VALUE` when used as an argument


dewiniaid:
I couldn't figure out how to succinctly word the summary.

When receiving an Enum as an argument, it's not possible to compare it against the class that defined it.  That is, `args['episode'] == Episode.NEWHOPE` will be `False` on `query { foo(episode:NEWHOPE) }`.  This seems counter-intuitive

```python
class Episode(Enum):
    NEWHOPE = 4
    EMPIRE = 5
    JEDI = 6


class Query(ObjectType):
    foo = String(
        args={'blah': Episode(default_value=Episode.NEWHOPE)}
    )

    def resolve_foo(self, args, context, info):
        import sys

        blah = args.get('blah')
        return (
            "Value of blah is: {!r}\n"
            "(blah == NEWHOPE) is: {!r}\n"
            "Python version is: {!r}\n"
            ""
        ).format(
            blah,
            blah == Episode.NEWHOPE,
            sys.version
        )

```

```
query {
  foo(blah:NEWHOPE)
}
```

yields (de-JSOnified for easier reading):

```
Value of blah is: 4
(blah == NEWHOPE) is: False
Python version is: '3.5.2 (default, Nov 17 2016, 17:05:23) \\n[GCC 5.4.0 20160609]'
```

"Native" Enums don't compare to their 'actual' representation either...

```python
>>> import enum
>>> class Episode(enum.Enum):
...     NEWHOPE = 4
...
>>> Episode.NEWHOPE == 4
False
```
... but they can be converted; e.g. `Episode(4) == Episode.NEWHOPE`, which does not work in Graphene.   Additionally, Python's IntEnum class does permit straight comparisons to integer values.

The behavior here appears to contradict the documentation (emphasis mine):

> graphene.Enum uses enum.Enum internally (or a backport if that’s not available) *and can be used in the exact same way*.

IMHO, the correct behavior here would be for resolve_foo() to get `Episode.NEWHOPE` rather than `4` for the argument here, but this presumably would break a lot of existing code.  Perhaps some sort of option to accomodate this?  

Alternatively, having a `graphene.IntEnum` that's functionally equivalent to `enum.IntEnum` would be nice.

PR #879, opened 2018-12-17
Correctly resolve Enum 

avivey:
See #877, #756, #444.

This changes introduces "new style enums" which are more natural to work with, having the value provided to the resolver is an enum value.
This only works with Python-Enum based enums (`graphene.Enum.from_enum()`) and is disabled by default (pass `legacy_enum_resolver=False` to enable).

There's still a bug with specifying default values to enum arguments, but that bug is in `graphql-core`, and there's no reasonable workaround in `graphene` for it. We'll need `graphql-core-next` to better handle enums.

In part, this change reverts #446 for new-style enums. #446 added a custom `__eq__` to the underlying python-enum backing graphene-enums. Some magic in the Python stack expects that if you provide `__eq__` you should also provide `__hash__` (Which makes sense), rendering the resulting enum instances un-hashable.
Since having `EnumInstance == EnumInstance.Value` doesn't make sense in a typed world (such as GraphQL), I prefer to remove this and resolve the underlying problem instead (provide resolvers with proper values).

### DIFF
--- a/examples/starwars/data.py
+++ b/examples/starwars/data.py
@@ -3,14 +3,14 @@ droid_data = {}
 
 
 def setup():
-    from .schema import Human, Droid
+    from .schema import Human, Droid, Episode
 
     global human_data, droid_data
     luke = Human(
         id="1000",
         name="Luke Skywalker",
         friends=["1002", "1003", "2000", "2001"],
-        appears_in=[4, 5, 6],
+        appears_in=[Episode.NEWHOPE, Episode.EMPIRE, Episode.JEDI],
         home_planet="Tatooine",
     )
 
@@ -18,7 +18,7 @@ def setup():
         id="1001",
         name="Darth Vader",
         friends=["1004"],
-        appears_in=[4, 5, 6],
+        appears_in=[Episode.NEWHOPE, Episode.EMPIRE, Episode.JEDI],
         home_planet="Tatooine",
     )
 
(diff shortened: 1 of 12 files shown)


avivey:
@ekampf

ekampf:
This is a breaking change.
So I guess this would require a version major\minor change...

@syrusakbary whats your policy on such changes?

ekampf:
@avivey let's discuss how we can make this backward-compatible.
It should work the old way by default until we make the cut to Python3...

allardhoeve:
What is needed for this to enter master? What is breaking about the change?

ekampf:
@allardhoeve the breaking change is that when . you have Enum arguments you'll now get the enum and not its value.
Say your enum is 
```
class Episode(graphene.Enum):
        NEWHOPE = 4
        EMPIRE = 5
        JEDI = 6
```

What you're getting in graphene now wuld be `Episode.NEWHOPE.value` and what you'll get after is `Episode.NEWHOPE`

We'll work to figure out how to make this backward compatible (have a flag somewhere indicating new\old enum support)

avivey:
@ekampf now supports old form by default.
New form only supported for the `from_enum()` form, because hopefully some day we can drop the `class Xx(Enum):` form and remove a lot of code from that.

phalt:
@dschaller @avivey we could bundle this over the graphene3 release?

pngnviko:
ping

ekampf:
There's no way to reopen PRs?  or is it a permission Im missing?

avivey:
I don't see an "open" button... maybe create a new branch from this commit
and create  new pr?

Issue #887, opened 2019-01-04
Document how to implement Connection fields and pagination


antoine-gallix:
From the [documentation on relay Connection](https://docs.graphene-python.org/en/latest/relay/connection/#connection), there is an example on how to define a connection object type, but in the part that shows how to define a connection field, it's not clear what should the resolver return. The example only show an empty list returned. What type of object is a connection resolver supposed to return? Could you show a concrete example?

Also the sentence in that paragraph is not clear:

> You can create connection fields in any Connection, in case any ObjectType that implements Node will have a default  Connection.

What does this sentence mean?

antoine-gallix:
What I found out:

The following worked to define a List field

```python

class Thing(SQLAlchemyObjectType):
    class Meta:
        model = models.Thing # sqlalchemy model

things = graphene.List(ThingGraphQLType)

@staticmethod
def resolve_things(args, info):
    # returning a query seems to be ok
    return Thing.get_query(info=info)
```

However the following does not:

```python
class ThingConnection(relay.Connection):
    class Meta:
        node = Thing

things = relay.ConnectionField(ThingConnection)

@staticmethod
def resolve_things(args, info):
    return Thing.get_query(info=info)
```

`graphql.error.located_error.GraphQLLocatedError: object of type 'BaseQuery' has no len()`

But this works:

```python
@staticmethod
def resolve_things(args, info):
    return Thing.get_query(info=info).all()
```

This is all based on try and error. That could be nice to explain that upfront in the doc.

antoine-gallix:
After the previous declaration, I see on my schema that ThingConnection exposes `first`,`last`,`before` and `after`. This makes me think that pagination is implemented. 

However if I try to use the `first` argument in my query, like `things(first:1) {edges{node{name}}}`, I get back an error like `graphql.error.located_error.GraphQLLocatedError: resolve_farm_connection() got an unexpected keyword argument 'first'`

Looks like it's not implemented. So I add the `first` argument to my resolver signature, without actually implementing the pagination:

```python
@staticmethod
resolve_things(args,info,first=None):
    # 'first' argument is not used anywhere, only added to the signature
    return models.Things.query().all()
```

There is three Things in my db. If I query `things {edges{node{name}}}`, I get 3 edges. Then if `things(first:1) {edges{node{name}}}`, I get only one. So after all even if I don't implement it myself, something in the Connection class will do it. But I still need to allow `first` in the resolver signature.

[...]

PR #892, opened 2019-01-17
doc:connection:precise expected return of connection resolver

antoine-gallix:
connection documentation lacks a lot of details: https://github.com/graphql-python/graphene/issues/887


antoine-gallix:
I wanted to improve the documentation because it was not clear to me what a resolver function was supposed to return. We managed to make it work by returning a list of objects, as no other way is documented. But as you seems to know better what the standard practice is, could you please explain it in detail so we can improve the documentation together?

phalt:
@ekampf can you suggest a better doc for this?

PR #910, opened 2019-02-27
Allow inheritance of the connection class

wakemaster39:
currently cannot inherit the connection class and supply a new meta class. Added support for this.

### DIFF
--- a/graphene/relay/connection.py
+++ b/graphene/relay/connection.py
@@ -50,8 +50,9 @@ class Connection(ObjectType):
         abstract = True
 
     @classmethod
-    def __init_subclass_with_meta__(cls, node=None, name=None, **options):
-        _meta = ConnectionOptions(cls)
+    def __init_subclass_with_meta__(cls, node=None, name=None, _meta=None, **options):
+        if not _meta:
+            _meta = ConnectionOptions(cls)
         assert node, "You have to provide a node in {}.Meta".format(cls.__name__)
         assert isinstance(node, NonNull) or issubclass(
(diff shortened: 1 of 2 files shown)


graphene/relay/connection.py
```diff
@@ -50,8 +50,9 @@ class Meta:
         abstract = True
 
     @classmethod
-    def __init_subclass_with_meta__(cls, node=None, name=None, **options):
-        _meta = ConnectionOptions(cls)
+    def __init_subclass_with_meta__(cls, node=None, name=None, _meta=None, **options):
+        if not _meta:
+            _meta = ConnectionOptions(cls)
```
keitheis: Maybe
```python
    def __init_subclass_with_meta__(cls, node=None, name=None, connection_options=None, **options):
        _meta = connection_options or ConnectionOptions(cls)
```
and have a test for this.

wakemaster39: I updated to add a unit test. I went with the _meta kwarg over connection_options because then it is the same name flowing through the entire process chain. 

If you think it makes more sense to change it to connection_options I will do that, I just think it feels weird to change the naming convention.

keitheis: I just looked at this function and made the common comment. It should be fine if this repo went with the convention.

wakemaster39:
I fixed the unit test. I also added the ability to have fields defined on the meta class before it is passed in along with the associated unit test

phalt:
@wakemaster39 some failing tests still.

PR #911, opened 2019-03-08
Fix build - Fix missing type annotations

ekampf:
`flake8` was Failing in Python 3.6 on not having `Type` and `Any` definitions.
These are special Python typing definitions so to fix this in Python3.6+ we need to:
`from typing import Type, Any`

In Python2, this is kind of meaningless so I just fake them to be
```
Type = type
Any = object
```

### DIFF
--- a/graphene/types/base.py
+++ b/graphene/types/base.py
@@ -1,5 +1,12 @@
+import six
+
 from ..utils.subclass_with_meta import SubclassWithMeta
 from ..utils.trim_docstring import trim_docstring
+
+if six.PY3:
+    from typing import Type
+else:
+    Type = type
 
 
 class BaseOptions(object):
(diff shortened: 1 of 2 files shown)


graphene/types/base.py
```diff
@@ -1,6 +1,13 @@
+import six
+
 from ..utils.subclass_with_meta import SubclassWithMeta
 from ..utils.trim_docstring import trim_docstring
 
+if six.PY3:
+    from typing import Type
+else:
```
dan98765: since the typing in this file is just type hints, which are just comments to the python interpreter, do you even need this for py27? What happens if you just remove the `else` completely and so only have Type defined for py3?

dan98765: I gave it a shot in https://github.com/graphql-python/graphene/pull/913, seems like omitting the `else` statement tests still pass.

ekampf: Didn't think of that... checking

ekampf: @dan98765 ok closing this PR in favor of yours

ekampf:
Closing in favor of #913

Issue #890, opened 2019-01-14
Incomplete example about query variables


hannes-petri-maxiv-lu-se:
In the docs, there's an example about how to feed variables to a query [(link)](https://docs.graphene-python.org/en/latest/execution/execute/)

However, the example is incomplete as there is nothing demonstrating how to consume the variables in the resolver:

    class Query(graphene.ObjectType):
        user = graphene.Field(User)

        def resolve_user(self, info):
            return info.context.get('user')

    schema = graphene.Schema(Query)
    result = schema.execute(
        '''query getUser($id: ID) {
            user(id: $id) {
                id
                firstName
                lastName
            }
        }''',
        variables={'id': 12},
    )

jkimbo:
Yes you're right. The code should be:

```python
class Query(graphene.ObjectType):
    user = graphene.Field(User, id=graphene.ID(required=True))

    def resolve_user(_, info, id):
        return get_user_by_id(id)

schema = graphene.Schema(Query)
result = schema.execute(
    '''query getUser($id: ID) {
        user(id: $id) {
            id
            firstName
            lastName
        }
    }''',
    variables={'id': 12},
)

**Comment by PsidomPC** [comment]
How can I use a customized class as `args` in `graphene.Field`?

```
class AdGroupConfigChangesArguments(graphene.InputObjectType):
    platform = graphene.String()
    platform_id = graphene.String()
    bid_strategy = graphene.String()
```
I tried with:

`graphene.Field(
    graphene.List(AdGroupConfigChanges), 
    args=AdGroupConfigChangesArguments, 
    resolver=adgroup_config_changes_resolver)`

But it gives an error:

> AssertionError: Arguments in a field have to be a mapping

PR #920, opened 2019-03-16
Update execution doc with correct way to use variables

jkimbo:
Fixes https://github.com/graphql-python/graphene/issues/890

### DIFF
--- a/docs/execution/execute.rst
+++ b/docs/execution/execute.rst
@@ -33,7 +33,7 @@ You can pass context to a query via ``context``.
 
 
 Variables
-_______
+_________
 
 You can pass variables to a query via ``variables``.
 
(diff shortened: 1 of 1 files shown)


docs/execution/execute.rst
```diff
@@ -33,18 +33,18 @@ You can pass context to a query via ``context``.
...
 
 .. code:: python
 
     class Query(graphene.ObjectType):
-        user = graphene.Field(User)
+        user = graphene.Field(User, id=graphene.ID(required=True))
 
-        def resolve_user(self, info):
-            return info.context.get('user')
+        def resolve_user(_, info, id):
```
dan98765: suggest keeping `self` here. It's python convention. Plus we have it in the example above this for `resolve_name`.

jkimbo: As per other discussions it looks like we're going to standardise on `root` for top level resolvers so I'll change this.

PR #921, opened 2019-03-16
Update object type docs

jkimbo:
Now that the default resolver handles both objects and dicts: https://github.com/graphql-python/graphene/pull/638

### DIFF
--- a/docs/types/objecttypes.rst
+++ b/docs/types/objecttypes.rst
@@ -57,8 +57,8 @@ so the first argument to the resolver method ``self`` (or ``root``) need
 not be an actual instance of the ``ObjectType``.
 
 If an explicit resolver is not defined on the ``ObjectType`` then Graphene will
-attempt to use a property with the same name on the object that is passed to the
-``ObjectType``.
+attempt to use a property with the same name on the object or dict that is
+passed to the ``ObjectType``.
 
 .. code:: python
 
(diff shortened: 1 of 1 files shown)


docs/types/objecttypes.rst
```diff
@@ -70,54 +70,18 @@ attempt to use a property with the same name on the object that is passed to the
...
-            default_resolver = dict_resolver
-
-        first_name = graphene.String()
-        last_name = graphene.String()
-
-    class Query(graphene.ObjectType):
-        me = graphene.Field(Person)
-
-        def resolve_me(_, info):
+        def resolve_best_friend(_, info):
```
dan98765: ahh I just opened an issue on your other PR about putting `self` into each method as per python convention, and also its used in other examples in the docs. But now here I see this place in the docs it is also set to be `_` previously. Having it both ways in the docs feels not-consistent.

In the case of this particular PR your addition here matches the convention in this file of not having `self` so I'm fine with just merging this in and maybe we can figure out the whole self-vs-underscore thing across the docs separately...

ProjectCheshire: And elsewhere it is root, info, **args

Consistent would be fantastic

ProjectCheshire: And elsewhere it is root, info, **args

Consistent would be fantastic

jkimbo: @dan98765 so the reason I don't use `self` when defining resolvers is because `self` means a specific thing in python (the instance of the current class) but in Graphene the first argument to the resolver is not an instance of the current class. Actually resolvers work like static functions and the first param is whatever the parent resolver returned.

Using `self` in the past has led to people being confused about what the value actually is and assuming it's the current instance (which is not unreasonable) so I've taken to naming the value as `_` when it's at the top of the tree (because if there is no `root` value then it's always `None`) or calling it a name that refers to what I'm expecting to receive. For example:

```python
class User(graphene.ObjectType):
	name = graphene.String()

	def resolve_name(user, info):  # <-- the first argument will be a Django model instance
		return user.name

	user_by_id = graphene.Field(User, id=graphene.Int())
	
	def resolve_user_by_id(_, info, id):  # <-- no root value so the first argument is always None
		return DjangoUser.objects.get(id=id)
```

I agree that we should standardise on something and actually thinking about it it makes sense to standardise on `root` for the top level Query resolvers (rather than `_`) so I'm happy to change the documentation to reflect that. I would also strongly recommend not using `self` for any resolver documentation because I think it ends up being confusing. Does that make sense @dan98765 @ProjectCheshire ?

dan98765: Thanks for the explanation, @jkimbo! I agree that we should not use `self` unless the it actually does represent the instance of the current class in the standard python way.

> Actually resolvers work like static functions and the first param is whatever the parent resolver returned.

Could resolvers be decorated with `@staticmethod` to make this more explicit? Or does that break something?

jkimbo: I'm unsure if you can add `@staticmethod` to resolvers actually but I don't see why it wouldn't work. It would get quite tedious to have to do it everywhere though.

Issue #870, opened 2018-11-26
Resolver Not Receiving Arguments when Nested


getglad:
I have a list of classes stored in memory that I am trying to parse through various types. It is referenced through the method `get_inventory()`.

When I call the classes individually, they resolve as I would expect.

But when I try to nest one in the other, the value is returning null.

The code, followed by some examples:

    class Account(graphene.ObjectType):
    
        account_name = graphene.String()
        account_id = graphene.String()
    
        def resolve_account(
            self, info,
            account_id=None,
            account_name=None
        ):
            
            inventory = get_inventory()
    
            result = [Account(
                account_id=i.account_id,
                account_name=i.account_name
            ) for i in inventory if (
                (i.account_id == account_id) or 
                (i.account_name == account_name)
            )]
    
            if len(result):
                return result[0]
            else:
                return Account()
    
    account = graphene.Field(
        Account, 
        resolver=Account.resolve_account, 
        account_name=graphene.String(default_value=None),
        account_id=graphene.String(default_value=None)
    )
    
    class Item(graphene.ObjectType):
    
        item_name = graphene.String()
        region = graphene.String()
        account = account
    
        def resolve_item(
            self, info,
            item_name=None
        ):
            inventory = get_inventory()
    
            result = [Item(
                item_name=i.item_name,
                region=i.region,
                account=Account(
                    account_id=i.account_id
                )
            ) for i in inventory if (
                (i.item_name == item_name)
            )]
    
            if len(result):
                return result[0]
            else:
                return Item()
    
    item = graphene.Field(
        Item, 
        resolver=Item.resolve_item, 
        item_name=graphene.String(default_value=None)
    )
    
    class Query(graphene.ObjectType):
        account = account
        item = item
    
    schema = graphene.Schema(query=Query)

Let's assume I have an account `foo` that has an item `bar`. The below queries return the fields correctly.

    {
      account(accountName:"foo") {
        accountName
        accountId
      }
    }
    
    {
      item(itemName: "bar") {
        itemName
        region
      }
    }

So if I wanted to find the account that has the item `bar`, I would think I could query `bar` and get `foo`. But it returns the `account` fields as `null`.

    {
      item(itemName: "bar") {
        itemName
        region
        account {
          accountId
          accountName
        }
      }
    }

Recall that as part of `resolve_item`, I am doing `account=Account(account_id=i.account_id)` - I would expect this to work.

If I alter the last return statement of `resolve_account` to the below, `accountId` always returns `yo`.

[...]

PR #930, opened 2019-03-28
Allow initialization of ObjectType with dynamic private state

getglad:
Proposing change to concept of "private state" to allow for more dynamic assignment.

While looking into patterns to pass data from parent -> child (https://github.com/graphql-python/graphene/issues/870), found that `ObjectType` could use `kwarg` prepended by an `_`.  At first thought there was a bug (https://github.com/graphql-python/graphene/blob/master/graphene/types/objecttype.py#L98) where `getattr` was missing a default of `None`, but later found in test case that may have been intentional (https://github.com/graphql-python/graphene/commit/2975e1fc42709cfd8c2331a57dd91091f6ad2ef4#diff-a76bb330091e6980d5fe6d9e5958a2edR78) to force explicit declaration.

From my view point, there are benefits to allow dynamic `kwarg`, especially for use cases with differentiated resolvers (ie, resolve one way when a parent node, another when a child, a third when a child of a child).

Either way, I think this capability should also be surfaced in the documentation. Happy to also take that on if we can establish the "graphenic" way of doing it.

### DIFF
--- a/graphene/tests/issues/test_870.py
+++ b/graphene/tests/issues/test_870.py
@@ -1,0 +1,80 @@
+import graphene
+
+
+class SecondChild(graphene.ObjectType):
+    hello = graphene.String()
+
+
+def resolve_second_child(self, info):
+    return SecondChild(hello=self.second_child._test)
+
+
+second_child_f = graphene.Field(SecondChild, resolver=resolve_second_child)
+
+
+class FirstChild(graphene.ObjectType):
+    hello = graphene.String()
+    second_child = second_child_f
+
+
+def resolve_first_child(self, info):
+    message = self.first_child._test + " stranger"
+
+    test_case = {"_test": message}
+
+    return FirstChild(
+        hello=self.first_child._test, second_child=SecondChild(**test_case)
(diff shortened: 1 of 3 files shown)


graphene/tests/issues/test_870.py
```diff
@@ -0,0 +1,80 @@
...
+
+class SecondChild(graphene.ObjectType):
+    hello = graphene.String()
+
+
+def resolve_second_child(self, info):
+    return SecondChild(hello=self.second_child._test)
+
+
+second_child_f = graphene.Field(SecondChild, resolver=resolve_second_child)
```
ekampf: Why do you define fields separately and not inside the class?
```
class FirstChild(graphene.ObjectType):
     hello = graphene.String()
     second_child = graphene.Field(SecondChild)

     def resolve_second_child(self, info):
        return SecondChild(hello=self.second_child._test)
```

getglad: I felt it made the steps more explicit, but if that's the preference/code style for the test, I can make the change

PR #934, opened 2019-04-03
Allow the Connection node to be wrapped in a `NonNull` type

minhtule:
When we declare the Connection `node` to be `NonNull` like below,
```python
class MyConnection(Connection):
    class Meta:
        node = NonNull(String)
```
we fail the assertion 
```python
assert issubclass(node, (Scalar, Enum, ObjectType, Interface, Union, NonNull))
```

However, this should be allowed according to the [Relay Connection specification](https://facebook.github.io/relay/graphql/connections.htm#sec-Node),
> An “Edge Type” must contain a field called node. This field must return either a Scalar, Enum, Object, Interface, Union, or a Non‐Null wrapper around one of those types.

Alternatively, we could define the `Edge` in the connection.
```python
class MyConnection(Connection):
    class Meta:
        node = String # We have another assertion that make this `Meta` field required!

    class Edge:
        node = NonNull(String)
```
However, from the assertion that `Meta`'s `node` field is required, `Edge`  isn't meant to be used to declare `node`.

### DIFF
--- a/graphene/relay/connection.py
+++ b/graphene/relay/connection.py
@@ -53,7 +53,7 @@ class Connection(ObjectType):
     def __init_subclass_with_meta__(cls, node=None, name=None, **options):
         _meta = ConnectionOptions(cls)
         assert node, "You have to provide a node in {}.Meta".format(cls.__name__)
-        assert issubclass(
+        assert isinstance(node, NonNull) or issubclass(
             node, (Scalar, Enum, ObjectType, Interface, Union, NonNull)
         ), ('Received incompatible node "{}" for Connection {}.').format(
             node, cls.__name__
--- a/graphene/relay/tests/test_connection.py
+++ b/graphene/relay/tests/test_connection.py
@@ -108,6 +108,17 @@ def test_edge_with_bases():
     assert edge_fields["other"].type == String
 
 
+def test_edge_with_nonnull_node():
+    class MyObjectConnection(Connection):
+        class Meta:
+            node = NonNull(MyObject)
+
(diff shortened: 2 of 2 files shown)


graphene/relay/connection.py
```diff
@@ -53,7 +53,7 @@ class Meta:
     def __init_subclass_with_meta__(cls, node=None, name=None, **options):
         _meta = ConnectionOptions(cls)
         assert node, "You have to provide a node in {}.Meta".format(cls.__name__)
-        assert issubclass(
+        assert isinstance(node, NonNull) or issubclass(
```
Nabellaleen: Is `NonNull` really the only case where the node is an instance instead of a subclass ? I'm not sure of this :)

minhtule: In theory, we should be able to use instances of `String`, `Int`,`ObjectType` etc. However, because we use this `node` declaration in a `Field` ([line 70](https://github.com/graphql-python/graphene/pull/934/files#diff-50d2eb86a71015fe93ccc0c11414760bR70)), it will throw an exception if we do
```python
class MyConnection(Connection):
    class Meta:
        node = String(required=True)
```

ekampf: Out of that list, `NonNull` is the only one we'll get as an instance.

ekampf: Also, @minhtule 's example of `node = String(required=True)` is invalid as we expect a Type to be passed there. We create the `Node` field internally (as it needs a special name etc).
The only valid non-type input is `node = NonNull(String)`

PR #935, opened 2019-04-05
Update aniso8601 dependency to allow versions 4+.

crunk1:
I accidentally updated my local aniso8601 to 6.0.0. My projects were running fine with graphene==2.1.3 and aniso8601==6.0.0. Later, I noticed the pip warnings telling me they were incompatible versions.

I ran tox against aniso8601 versions 4.0.0, 5.0.0, and 6.0.0 and compared against the results from running against 3.0.2. Results look the same at a glance.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -88,7 +88,7 @@ setup(
         "six>=1.10.0,<2",
         "graphql-core>=2.1,<3",
         "graphql-relay>=0.4.5,<1",
-        "aniso8601>=3",
+        "aniso8601>=3,<=6.0.*",
     ],
     tests_require=tests_require,
     extras_require={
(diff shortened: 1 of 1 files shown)


ekampf:
Why not just update to `~=6.0.0` ?

crunk1:
To not break people on aniso8601 3.x-5.x

ekampf:
@crunk1 it feels wrong. Saying any version`>=3` is just like saying no version dependency at all...
`aniso8601` might break something in version 7 and we wont know...
I'd rather have `>=3, <=6.0.0` to be safe

crunk1:
Updated. I don't know what your thoughts are on allowing future compatibility changes, but take a look.

helloqiu:
Hi there.I received the same warning as @crunk1 and I found this PR.    
Do the maintainers of this project have plans on using apps like [renovate](https://github.com/apps/renovate) for automated dependencies updating? Renovate can make PRs automatically as long as the upstream publish new versions, and tests and changelog can tell whether the updates may break graphene.

dvndrsn:
Pyup is nice too: https://pyup.io/

PR #947, opened 2019-04-21
Change asserts to manually raising AssertionError

dsanders11:
See graphql-python/graphql-core#244 for discussion of why this change is needed. The issue exists in several `graphql-python` projects (`graphql-core`, `graphene`, `graphene-django`), but biting this one off first as it was smaller than the changes needed in `graphql-core`.

Ideally the usage of `AssertionError` would be dropped in favor of `TypeError` and `ValueError`, but that would have to wait for a new major version release as it's a backwards incompatible change.

### DIFF
--- a/graphene/relay/connection.py
+++ b/graphene/relay/connection.py
@@ -52,12 +52,18 @@ class Connection(ObjectType):
     @classmethod
     def __init_subclass_with_meta__(cls, node=None, name=None, **options):
         _meta = ConnectionOptions(cls)
-        assert node, "You have to provide a node in {}.Meta".format(cls.__name__)
-        assert isinstance(node, NonNull) or issubclass(
+        if not node:
+            raise AssertionError(
+                "You have to provide a node in {}.Meta".format(cls.__name__)
+            )
+        if not isinstance(node, NonNull) and not issubclass(
             node, (Scalar, Enum, ObjectType, Interface, Union, NonNull)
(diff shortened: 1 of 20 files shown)


ekampf:
As I mentioned in the issue Im not sure this is the correct approach avoiding `assert` statements all together,
What I am sure is that *manually* throwing AssertionError is wrong.
If we're converting assertion checks to runtime checks we should use the appropriate errors (like `ValueError` etc)

ekampf:
I also wrote this on the issue but I think the PR is too big - removing *all* uses of `assert` regardless of case.
I think the change is correct in the `Date(Scalar)` (but should throw the right error and not `AssertionError`) but I'm not sure about a lot of the other places...

(if you want to separate the changes to `Date` and similar changes to a separate PR I would gladly approve that one and we can discuss the rest case by case)

dsanders11:
> If we're converting assertion checks to runtime checks we should use the appropriate errors (like `ValueError` etc)

That would require a new major version of `graphql-core`, `graphene`, `django-graphene`, etc. Changing the exception type from `AssertionError` changes the contract of the function, so you can't change that and not potentially break other people's code. For example, the code I mentioned in `django-graphene` at [this line](https://github.com/graphql-python/graphene-django/blob/master/graphene_django/converter.py#L35) depends on catching an `AssertionError`.

Issue #951, opened 2019-04-25
Confusing documentation on self argument of mutate method of Mutations


allardhoeve:
Hi there,

The documentation on Mutations is a little confusing about the arguments the `mutate` method of a `Mutation` receives. Two examples list the method as:

```python
# https://github.com/allardhoeve/graphene/blame/abff3d75/docs/types/mutations.rst#L22

class Something(Mutation):
    def mutate(self, info, *args):
        pass
```

This is actually incorrect, as the `mutate` method is always a `@staticmethod`, because the resolver is unbound [here in graphene/types/mutation.py](https://github.com/graphql-python/graphene/commit/2d557d6ed7290#diff-bc71456af798aee511eb900d021e3978R44).

This is documented further on as:

```python
# https://github.com/allardhoeve/graphene/blame/abff3d75a39/docs/types/mutations.rst#L107

class CreatePerson(graphene.Mutation):
    ...
    @staticmethod
    def mutate(root, info, person_data=None):
        pass
```

I would like to document this better in this part of the documentation, because it has caused me a lot of time to figure this out and would like to spare other that effort.

Can you elaborate on when `root` is set and what it will contain? I've only seen cases where `root` is `None`.

Also, can you elaborate on why you chose to make the resolver method `mutate` a static method by default? It being a static method makes it hard to implement inheritance  between `Mutation` classes.

Finally, how would I go about having a resolver method that is NOT a static method if I wanted to? The code [suggests](https://github.com/graphql-python/graphene/blob/abff3d75a39/graphene/types/mutation.py#L63) I could use `Meta` attributes to do so, but the documentation does not tell me how. I would like to add that to the documentation.

Edit: The resolvers always being a `staticmethod` is documented [here](https://docs.graphene-python.org/en/latest/types/objecttypes/), but you have to first know that `def mutate` is considered a resolver and also you'd have to really remember that.

Thanks

dvndrsn:
Hey @allardhoeve, thanks for your question!

The first argument of `mutate`, `mutate_and_get_payload`, and `resolve_xxx` method is always `root` instead of `self`, they are treated in the same way as a `@staticmethod` decorator. The documentation can definitely be more clear here. When you're resolving a query, `root` intuitively makes sense as the value object retuned from the last resolver you visited in the graph of results that is being used to resolve the current type's fields.

[...]

PR #952, opened 2019-04-25
Modify documentation to be more clear about resolvers being static

allardhoeve:
See #951

### DIFF
--- a/docs/execution/dataloader.rst
+++ b/docs/execution/dataloader.rst
@@ -111,8 +111,10 @@ leaner code and at most 4 database requests, and possibly fewer if there are cac
         best_friend = graphene.Field(lambda: User)
         friends = graphene.List(lambda: User)
 
-        def resolve_best_friend(self, info):
-            return user_loader.load(self.best_friend_id)
+        @staticmethod
+        def resolve_best_friend(root, info):
+            return user_loader.load(root.best_friend_id)
 
-        def resolve_friends(self, info):
-            return user_loader.load_many(self.friend_ids)
+        @staticmethod
+        def resolve_friends(root, info):
+            return user_loader.load_many(root.friend_ids)
--- a/docs/execution/dataloader.rst
+++ b/docs/execution/dataloader.rst
@@ -112,9 +112,9 @@ leaner code and at most 4 database requests, and possibly fewer if there are cac
         friends = graphene.List(lambda: User)
 
         @staticmethod
-        def resolve_best_friend(root, info):
+        def resolve_best_friend(user, info):
             return user_loader.load(root.best_friend_id)
 
         @staticmethod
-        def resolve_friends(root, info):
+        def resolve_friends(user, info):
             return user_loader.load_many(root.friend_ids)
--- a/docs/relay/connection.rst
+++ b/docs/relay/connection.rst
@@ -41,5 +41,6 @@ that implements ``Node`` will have a default Connection.
         name = graphene.String()
         ships = relay.ConnectionField(ShipConnection)
 
-        def resolve_ships(self, info):
+        @staticmethod
+        def resolve_ships(root, info):
             return []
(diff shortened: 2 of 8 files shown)


docs/execution/dataloader.rst
```diff
@@ -111,8 +111,10 @@ leaner code and at most 4 database requests, and possibly fewer if there are cac
         best_friend = graphene.Field(lambda: User)
         friends = graphene.List(lambda: User)
 
-        def resolve_best_friend(self, info):
-            return user_loader.load(self.best_friend_id)
+        @staticmethod
+        def resolve_best_friend(root, info):
```
jkimbo:
```suggestion
        def resolve_best_friend(user, info):
```

`root` has a specific meaning in graphql (it’s the value that is passed into the root of the tree) and so anything that is it the `Query` object should use variable names that indicate what the variable is, so `user` in this case.

docs/execution/dataloader.rst
```diff
@@ -111,8 +111,10 @@ leaner code and at most 4 database requests, and possibly fewer if there are cac
...
+            return user_loader.load(root.best_friend_id)
 
-        def resolve_friends(self, info):
-            return user_loader.load_many(self.friend_ids)
+        @staticmethod
+        def resolve_friends(root, info):
```
jkimbo:
```suggestion
        def resolve_friends(user, info):
```

docs/relay/connection.rst
```diff
@@ -41,5 +41,6 @@ that implements ``Node`` will have a default Connection.
         name = graphene.String()
         ships = relay.ConnectionField(ShipConnection)
 
-        def resolve_ships(self, info):
+        @staticmethod
+        def resolve_ships(root, info):
```
jkimbo:
```suggestion
        def resolve_ships(faction, info):
```

allardhoeve:
> @allardhoeve decorating resolvers with @staticmethod is not necessary but thanks for the change from self

Good suggestions, I'll apply them next week and then re-request the merge.

I disagree about the necessity of `@staticmethod` though. I understand that you unbind the method in the resolver code, so technically for GraphQL it is not needed. But there are other things to consider besides this.

1. IDEs such as PyCharm will warn that the method, because it is still a method even though GraphQL unbinds it, should be marked as such.
2. These IDEs will also warn you that coding standards indicate that the first argument to a method is normally `self`.

Of course you can ignore these standards, but they are confusing to code reviewers and to programmers new to GraphQL, such as myself.

Lastly the methods are no longer callable from the outside of GraphQL without also using the unbind magic, because the signature no longer matches a method, because the `self` argument is not in the signature. This would prevent you from calling the resolver properly, for example in unittests.

I feel that the documentation for GraphQL should be as clear as possible and should reflect this rather odd unbinding of the resolver well, so new programmers are sure about the interface.

You can of course always leave the decorator off you own functions if you want. To me this is a documentation issue.

ProjectCheshire:
I disagree on needing static also because resolvers do not need to be defined next to the fields, but may be defined outside of, and may very well be for more complex ones. 

In that case, adding staticmethod makes it confusing. And doing it only for resolvers when defined with the field makes it inconsistent and confusing.

allardhoeve:
Indeed, not every resolver needs to be a staticmethod. If you use an outside function as a resolver, then no, you definitely shouldn't use a decorator. But as defined in a class, they probably should be marked as such, as they are essentially functions after becoming unbound by GraphQL.

My request is not at all to force everyone to use static method decorators everywhere, but just that we improve the documentation on resolvers better, because they are inconsistently documented currently.

My suggestion is still that for resolvers that are defined on NodeType objects, we still document them as static methods, as they are. And the part in the documentation on resolvers is easily missed.

But the situation improves with changing `self` with the subject of the function such as `ship`, thanks!

phalt:
Is it possible to add a documentation note about this issue with some IDEs and leave the default examples without? Either way, we should progress this PR or close it.

PR #962, opened 2019-05-07
Add contributing.MD, makefile commands

phalt:
Shuffle around the contributing file to a separate section.
Use makefile commands to make set up easier.


CONTRIBUTING.md
```diff
@@ -0,0 +1,57 @@
...
+After cloning this repo, create a [virtualenv](https://virtualenv.pypa.io/en/stable/) and ensure dependencies are installed by running:
+
+```sh
+virtualenv venv
+source venv/bin/activate
+make dev-setup
+git checkout -B YOUR_BRANCH_NAME
+```
+
+You are not ready to start developing
```
mvanlonden: Why did I do this then? 😛

CONTRIBUTING.md
```diff
@@ -0,0 +1,57 @@
+# Contributing
+
+## Set up
+
+After cloning this repo, create a [virtualenv](https://virtualenv.pypa.io/en/stable/) and ensure dependencies are installed by running:
```
mvanlonden: Could we offer [pipenv](https://docs.pipenv.org/en/latest/) as a lighter weight alternative?

CONTRIBUTING.md
phalt: Hah I just copy/pasted the original docs

CONTRIBUTING.md
phalt: @mvanlonden from my personal experience the packaging story of Python is in flux at the moment, I'd rather pick the basic one and when things settle down and a clear winner appears we move over to use it.

dvndrsn: Agree on Pipenv. For personal projects, it works fine and has a lot of nice convenience features but there are some things that make it less desirable for cross platform development and especially building libraries.

ekampf: "lighter weight" ?  what's wrong with `virtualenv` ?

Makefile
```diff
@@ -9,3 +9,12 @@ docs:
 	pip install -r requirements.txt &&\
 	make html &&\
 	cd -
+
+dev-setup:
+	pip install -e ".[test]"
```
jkimbo: These commands are going to conflict with https://github.com/graphql-python/graphene/pull/971/files#diff-b67911656ef5d18c4ae36cb6741b7965

dvndrsn:
@jkimbo i was just following @phalt 's lead with those conflicting commands, no worries. Either works.

phalt:
I'm happy to close this in favour of the other one.

jkimbo:
@phalt I think the CONTRIBUTING doc is still really useful from this PR and I didn't mean for you to close it! I was just pointing out that 2 separate PRs have added makefile commands that are doing the same thing (but with different names).

PR #971, opened 2019-05-22
expose livehtml autobuild in Makefile + Add API autodoc

dvndrsn:
Autobuild was previously set up in Makefile, but did not have its requirements frozen in `requirements.txt` and the command was not documented well. Added API level documentation for parameters and arguments for the core graphene objects.

- Added `make docs-live` target to the root project Makefile. This starts a live http server which refreshes with docs changes.
- Added `sphinx-autobuild ` to documentation requirements.
- Added `autodoc` configuration to generate documentation from classes automatically.
- API documentation attributes and parameters for core graphene objects and interfaces.
- General tidying and documentation.
- Fixed a flakey test involving `datetime.now()` by setting static values for times in tests.

PR #975, opened 2019-05-24
Typo in documentation

mikebobadilla:
Adds missing word in docs.

 schema will the same as the class ->  schema will be the same as the class


dvndrsn:
It looks like your CI run failed due to a flakey test using datetime. I fixed this in another branch in this patch: https://github.com/graphql-python/graphene/pull/971/commits/fa3ecce9e3e279f580ad975148aa0cf5c3a4cae4

PR #984, opened 2019-06-01
Refactor for f strings.

changeling:
Left the ``deprecated.py` alone for now, but it perhaps that deserves re-factoring for a very minor performance win?

Skipping some of `relay/tests/test_connection_query.py` for nested quoting laziness on my part.

### DIFF
--- a/docs/relay/nodes.rst
+++ b/docs/relay/nodes.rst
@@ -52,7 +52,7 @@ Example of a custom node:
 
         @staticmethod
         def to_global_id(type, id):
-            return '{}:{}'.format(type, id)
+            return f'{type}:{id}'
 
         @staticmethod
         def get_node_from_global_id(info, global_id, only_type=None):
--- a/graphene/types/base.py
+++ b/graphene/types/base.py
@@ -20,10 +20,10 @@ class BaseOptions(object):
         if not self._frozen:
             super(BaseOptions, self).__setattr__(name, value)
         else:
-            raise Exception("Can't modify frozen Options {}".format(self))
+            raise Exception(f"Can't modify frozen Options {self}")
 
     def __repr__(self):
-        return "<{} name={}>".format(self.__class__.__name__, repr(self.name))
+        return f"<{self.__class__.__name__} name={repr(self.name)}>"
 
 
 class BaseType(SubclassWithMeta):
--- a/graphene/types/tests/test_structures.py
+++ b/graphene/types/tests/test_structures.py
@@ -19,7 +19,7 @@ def test_list_with_unmounted_type():
 
     assert (
         str(exc_info.value)
-        == "List could not have a mounted String() as inner type. Try with List(String)."
+        == "List can not have a mounted String() as inner type. Try List(String)."
     )
 
 
@@ -97,7 +97,7 @@ def test_nonnull_with_unmounted_type():
 
     assert (
         str(exc_info.value)
-        == "NonNull could not have a mounted String() as inner type. Try with NonNull(String)."
+        == "NonNull can not have a mounted String() as inner type. Try NonNull(String)."
     )
 
 
--- a/UPGRADE-v2.0.md
+++ b/UPGRADE-v2.0.md
@@ -379,10 +379,7 @@ class Base(ObjectType):
     id = ID()
 
     def resolve_id(self, info):
-        return "{type}_{id}".format(
-            type=self.__class__.__name__,
-            id=self.id
-        )
+        return f"{self.__class__.__name__}_{self.id}"
 ```
 
 ### UUID Scalar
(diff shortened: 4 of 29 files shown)


ekampf:
Your `black` pre-commit is failing

ekampf:
Also please rebase from `next/master`

graphene/types/base.py
```diff
@@ -20,10 +20,10 @@ def __setattr__(self, name, value):
         if not self._frozen:
             super(BaseOptions, self).__setattr__(name, value)
         else:
-            raise Exception("Can't modify frozen Options {}".format(self))
+            raise Exception(f"Can't modify frozen Options {self}")
 
     def __repr__(self):
-        return "<{} name={}>".format(self.__class__.__name__, repr(self.name))
+        return f"<{self.__class__.__name__} name={repr(self.name)}>"
```
ekampf: why `repr(self.name)` and not just `self.name!r` ?

dvndrsn:
F-Strings are not supported in Python 2.7, so this PR would need to be merged in 3.0 when 2.7 and 3.4 support is dropped. Making sure this is tagged appropriately.

changeling:
> F-Strings are not supported in Python 2.7, so this PR would need to be merged in 3.0 when 2.7 and 3.4 support is dropped. Making sure this is tagged appropriately.

Good note, and thanks. I'd not thought to do that, as it's targeted at `next/master`, which is 3 only, but I'll do that going forward.

changeling:
> Per our discussion on slack, lets use this opportunity to consistently use either
> `f"{repr(value)}` or f"{value!r}".
> We decided on `repr` in Slack...

@ekampf, "View changes" doesn't offer changes, so I'm unable to mark this change request as resolved.

docs/relay/nodes.rst
```diff
@@ -52,7 +52,7 @@ Example of a custom node:
 
         @staticmethod
         def to_global_id(type, id):
-            return '{}:{}'.format(type, id)
+            return f'{type}:{id}'
```
jkimbo: Can we get black to run on the examples in the docs as well? In this case the quotes should be double quotes.

graphene/types/tests/test_structures.py
```diff
@@ -19,7 +19,7 @@ def test_list_with_unmounted_type():
 
     assert (
         str(exc_info.value)
-        == "List could not have a mounted String() as inner type. Try with List(String)."
+        == "List can not have a mounted String() as inner type. Try List(String)."
```
jkimbo:
```suggestion
        == "List cannot have a mounted String() as inner type. Try List(String)."
```

graphene/types/tests/test_structures.py
```diff
@@ -97,7 +97,7 @@ def test_nonnull_with_unmounted_type():
 
     assert (
         str(exc_info.value)
-        == "NonNull could not have a mounted String() as inner type. Try with NonNull(String)."
+        == "NonNull can not have a mounted String() as inner type. Try NonNull(String)."
```
jkimbo:
```suggestion
        == "NonNull cannot have a mounted String() as inner type. Try NonNull(String)."
```

graphene/types/tests/test_structures.py
changeling: I'll defer to the more common usage here. :) `cannot` it is.

Also changing in docstrings in `graphene/pyutils/signature.py`, lines 811 and 818.

docs/relay/nodes.rst
changeling: I've been trying to blacken them all, but clearly missed some. Lemme know if there are others. I'll do another run through.

changeling: NM. This is addressed in https://github.com/graphql-python/graphene/pull/992

changeling:
This is gonna need to wait until we decide on 3.5 support.

changeling:
Lookd like we have concensus on 3.6+, so this is ready to go (or review more)!

changeling:
What's the status here? Requested changes made, though the system still shows `Changes requested`. I believe that's your https://github.com/graphql-python/graphene/pull/984#pullrequestreview-244607933, @ekampf.

Can this be merged/reviewed so I can clear this from my end?

changeling:
Thoughts on reviewing/merging this, @ProjectCheshire, @ekampf, @dan98765 ?

KingDarBoja:
Any updates on this PR? Just curious as the author has no activity since July 2019 😨

ekampf:
@KingDarBoja its kind of stuck on conflicts it seem...  @changeling  ?

syrusakbary:
Apologizes for the delay actioning this PR.

This PR is taking the next steps to replace f-strings: https://github.com/graphql-python/graphene/pull/1158

PR #988, opened 2019-06-02
Change dependency to graphql-core-next

ekampf:


setup.py
```diff
@@ -80,15 +80,15 @@ def run_tests(self):
...
+        "graphql-relay>=3.0.0a0,<4",
+        "aniso8601~=6.0.0",
     ],
     tests_require=tests_require,
     extras_require={
         "test": tests_require,
-        "django": ["graphene-django"],
-        "sqlalchemy": ["graphene-sqlalchemy"],
+        # "django": ["graphene-django"],
+        # "sqlalchemy": ["graphene-sqlalchemy"],
```
jkimbo: Should these be removed?

Cito: I think so. It's much better and simpler to install graphene-django/sqlalchemy which then require a matching version of graphene. Also, it's unclear why graphene-django gets that extra treatment while graphene-flask does not. Better to keep graphene framework angostic.

Issue #970, opened 2019-05-20
Make example code more consistent in documentation


dvndrsn:
This might be a bit further off, as there are other things that need addressing in documentation-land, but it would be good to make the example code in the documentation more consistent by extracting examples from the same schema.

This example code could be hosted somewhere ([like glitch?][1]) and interactive queries could be added to the documentation using [sphinx-graphiql][2].

[1]: https://glitch.com/edit/#!/flask-hello-world
[2]: https://github.com/hasura/sphinx-graphiql

PR #992, opened 2019-06-04
[WIP] Initial work refactoring docs/ examples/ `class Meta` to class arguments, implementing explicit imports.

changeling:
Refactoring firehose imports to explicit imports in docs/, examples/.
Some blackening of example code in docs/.
Refactoring of `class Meta` into class arguments.

This is a first pass of docs/ and examples/ using explicit imports and refactoring `class Meta` into class arguments.

Very little thought towards functionality of the documentation examples, just standardizing.

### DIFF
--- a/docs/execution/execute.rst
+++ b/docs/execution/execute.rst
@@ -7,8 +7,8 @@ For executing a query a schema, you can directly call the ``execute`` method on 
 
 .. code:: python
 
-    schema = graphene.Schema(...)
-    result = schema.execute('{ name }')
+    schema = Schema(...)
+    result = schema.execute("{ name }")
 
 ``result`` represents the result of execution. ``result.data`` is the result of executing the query, ``result.errors`` is ``None`` if no errors occurred, and is a non-empty list if an error occurred.
 
@@ -21,14 +21,15 @@ You can pass context to a query via ``context``.
 
 .. code:: python
 
-    class Query(graphene.ObjectType):
-        name = graphene.String()
+    class Query(ObjectType):
+        name = String()
 
         def resolve_name(root, info):
-            return info.context.get('name')
+            return info.context.get("name")
 
-    schema = graphene.Schema(Query)
-    result = schema.execute('{ name }', context={'name': 'Syrus'})
+
+    schema = Schema(Query)
+    result = schema.execute("{ name }", context={"name": "Syrus"})
 
 
 
@@ -40,22 +41,23 @@ You can pass variables to a query via ``variables``.
 
 .. code:: python
 
-    class Query(graphene.ObjectType):
-        user = graphene.Field(User, id=graphene.ID(required=True))
+    class Query(ObjectType):
+        user = Field(User, id=ID(required=True))
 
         def resolve_user(root, info, id):
             return get_user_by_id(id)
 
-    schema = graphene.Schema(Query)
+
+    schema = Schema(Query)
     result = schema.execute(
-        '''
-          query getUser($id: ID) {
+        """
+            query getUser($id: ID) {
             user(id: $id) {
-              id
-              firstName
-              lastName
+                id
+                firstName
+                lastName
             }
-          }
-        ''',
-        variables={'id': 12},
+            }
+        """,
+        variables={"id": 12},
     )
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -12,7 +12,7 @@ Let’s build a basic GraphQL schema from scratch.
 Requirements
 ------------
 
--  Python (2.7, 3.4, 3.5, 3.6, pypy)
+-  Python (3.6+, pypy)
 -  Graphene (2.0)
 
 Project setup
@@ -35,15 +35,17 @@ one field: ``hello`` and an input name. And when we query it, it should return `
 
 .. code:: python
 
-    import graphene
+    from graphene import ObjectType, Schema, String
 
-    class Query(graphene.ObjectType):
-        hello = graphene.String(argument=graphene.String(default_value="stranger"))
+
+    class Query(ObjectType):
+        hello = String(argument=String(default_value="stranger"))
 
         def resolve_hello(self, info, argument):
-            return 'Hello ' + argument
+            return f"Hello {argument}"
 
-    schema = graphene.Schema(query=Query)
+
+    schema = Schema(query=Query)
 
 Querying
 --------
@@ -52,11 +54,11 @@ Then we can start querying our schema:
 
 .. code:: python
 
-    result = schema.execute('{ hello }')
-    print(result.data['hello']) # "Hello stranger"
+    result = schema.execute("{ hello }")
+    print(result.data["hello"])  # "Hello stranger"
 
     # or passing the argument in the query
     result = schema.execute('{ hello (argument: "graph") }')
-    print(result.data['hello']) # "Hello graph"
+    print(result.data["hello"])  # "Hello graph"
 
 Congrats! You got your first graphene schema working!
--- a/docs/relay/mutations.rst
+++ b/docs/relay/mutations.rst
@@ -12,13 +12,12 @@ subclass of ``relay.ClientIDMutation``.
 .. code:: python
 
     class IntroduceShip(relay.ClientIDMutation):
-
         class Input:
-            ship_name = graphene.String(required=True)
-            faction_id = graphene.String(required=True)
+            ship_name = String(required=True)
+            faction_id = String(required=True)
 
-        ship = graphene.Field(Ship)
-        faction = graphene.Field(Faction)
+        ship = Field(Ship)
+        faction = Field(Faction)
 
         @classmethod
         def mutate_and_get_payload(cls, root, info, **input):
@@ -28,8 +27,6 @@ subclass of ``relay.ClientIDMutation``.
             faction = get_faction(faction_id)
             return IntroduceShip(ship=ship, faction=faction)
 
-
-
 Accepting Files
 ---------------
 
@@ -37,13 +34,13 @@ Mutations can also accept files, that's how it will work with different integrat
 
 .. code:: python
 
-    class UploadFile(graphene.ClientIDMutation):
-         class Input:
-             pass
-             # nothing needed for uploading file
+    class UploadFile(ClientIDMutation):
+        class Input:
+            pass
+            # nothing needed for uploading file
 
-         # your return fields
-         success = graphene.String()
+        # your return fields
+        success = String()
 
         @classmethod
         def mutate_and_get_payload(cls, root, info, **input):
--- a/docs/relay/nodes.rst
+++ b/docs/relay/nodes.rst
@@ -14,12 +14,10 @@ Example usage (taken from the `Starwars Relay example`_):
 
 .. code:: python
 
-    class Ship(graphene.ObjectType):
-        '''A ship in the Star Wars saga'''
-        class Meta:
-            interfaces = (relay.Node, )
+    class Ship(ObjectType, interfaces=(relay.Node,)):
+        """A ship in the Star Wars saga"""
 
-        name = graphene.String(description='The name of the ship.')
+        name = String(description="The name of the ship.")
 
         @classmethod
         def get_node(cls, info, id):
@@ -45,26 +43,22 @@ Example of a custom node:
 
 .. code:: python
 
-    class CustomNode(Node):
-
-        class Meta:
-            name = 'Node'
-
+    class CustomNode(Node, name="Node"):
         @staticmethod
         def to_global_id(type, id):
-            return '{}:{}'.format(type, id)
+            return "{}:{}".format(type, id)
 
         @staticmethod
         def get_node_from_global_id(info, global_id, only_type=None):
-            type, id = global_id.split(':')
+            type, id = global_id.split(":")
             if only_type:
                 # We assure that the node type that we want to retrieve
                 # is the same that was indicated in the field type
-                assert type == only_type._meta.name, 'Received not compatible node.'
+                assert type == only_type._meta.name, "Received not compatible node."
 
-            if type == 'User':
+            if type == "User":
                 return get_user(id)
-            elif type == 'Photo':
+            elif type == "Photo":
                 return get_photo(id)
 
 
@@ -94,7 +88,7 @@ Example usage:
 
 .. code:: python
 
-    class Query(graphene.ObjectType):
+    class Query(ObjectType):
         # Should be CustomNode.Field() if we want to use our custom Node
         node = relay.Node.Field()
 
--- a/docs/types/enums.rst
+++ b/docs/types/enums.rst
@@ -11,9 +11,10 @@ You can create an ``Enum`` using classes:
 
 .. code:: python
 
-    import graphene
+    from graphene import Enum
 
-    class Episode(graphene.Enum):
+
+    class Episode(Enum):
         NEWHOPE = 4
         EMPIRE = 5
         JEDI = 6
@@ -22,7 +23,7 @@ But also using instances of Enum:
 
 .. code:: python
 
-    Episode = graphene.Enum('Episode', [('NEWHOPE', 4), ('EMPIRE', 5), ('JEDI', 6)])
+    Episode = Enum("Episode", [("NEWHOPE", 4), ("EMPIRE", 5), ("JEDI", 6)])
 
 Value descriptions
 ------------------
@@ -32,7 +33,7 @@ needs to have the ``description`` property on it.
 
 .. code:: python
 
-    class Episode(graphene.Enum):
+    class Episode(Enum):
         NEWHOPE = 4
         EMPIRE = 5
         JEDI = 6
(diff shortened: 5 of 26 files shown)


dvndrsn:
Hm.. This is gonna conflict quite a bit with my docs PR.. we need to start merging these faster.. 🤔

examples/starwars_relay/schema.py
```diff
@@ -1,34 +1,27 @@
-import graphene
+from graphene import ObjectType, Schema, Field, String
 from graphene import relay
```
dvndrsn: Why keep this as a module import rather than importing `relay` classes from `graphene` as well (all are available in the top level module) or import the classes from `graphene.relay`?

docs/relay/mutations.rst
```diff
@@ -28,22 +27,20 @@ subclass of ``relay.ClientIDMutation``.
...
 .. code:: python
 
-    class UploadFile(graphene.ClientIDMutation):
-         class Input:
-             pass
-             # nothing needed for uploading file
+    class UploadFile(ClientIDMutation):
+        class Input:
+            pass
+            # nothing needed for uploading file
```
dvndrsn: This comment is unrelated to the changes that you did:

It feels weird to have example code with commented out bits. I'd prefer to have a viable example which uses comments to highlight the important things.

docs/relay/mutations.rst
```diff
@@ -12,13 +12,12 @@ subclass of ``relay.ClientIDMutation``.
 .. code:: python
 
     class IntroduceShip(relay.ClientIDMutation):
-
         class Input:
-            ship_name = graphene.String(required=True)
```
dvndrsn: I think I like this better without `graphene graphene graphene` everywhere!

Did we have a discussion with folks to confirm that everyone is generally on the same page? The current convention is followed in all kinds of places so it will take some doing to change all the code examples.

docs/quickstart.rst
```diff
@@ -12,7 +12,7 @@ Let’s build a basic GraphQL schema from scratch.
 Requirements
 ------------
 
--  Python (2.7, 3.4, 3.5, 3.6, pypy)
+-  Python (3.6+, pypy)
```
dvndrsn: 3.5 should be supported too (until 3.8 is released in October).

dvndrsn: I guess we might want to clarify the versions supported and the plan in ROADMAP.md

docs/relay/nodes.rst
```diff
@@ -14,12 +14,10 @@ Example usage (taken from the `Starwars Relay example`_):
 
 .. code:: python
 
-    class Ship(graphene.ObjectType):
-        '''A ship in the Star Wars saga'''
-        class Meta:
-            interfaces = (relay.Node, )
+    class Ship(ObjectType, interfaces=(relay.Node,)):
```
dvndrsn: Do you prefer this way of supplying meta args? I've not used it much myself since I'm used to Meta inner class with Django.

docs/quickstart.rst
```diff
@@ -35,15 +35,17 @@ one field: ``hello`` and an input name. And when we query it, it should return `
...
 
-    class Query(graphene.ObjectType):
-        hello = graphene.String(argument=graphene.String(default_value="stranger"))
+
+    class Query(ObjectType):
+        hello = String(argument=String(default_value="stranger"))
 
         def resolve_hello(self, info, argument):
-            return 'Hello ' + argument
+            return "Hello " + argument
```
jkimbo: f string here?

docs/types/objecttypes.rst
```diff
@@ -18,15 +18,16 @@ This example model defines a Person, with a first and a last name:
...
-        full_name = graphene.String()
+
+    class Person(ObjectType):
+        first_name = String()
+        last_name = String()
+        full_name = String()
 
         def resolve_full_name(root, info):
-            return '{} {}'.format(root.first_name, root.last_name)
+            return "{} {}".format(root.first_name, root.last_name)
```
jkimbo:
```suggestion
            return f"{person.first_name} {person.last_name}"
```

docs/types/mutations.rst
```diff
@@ -38,20 +39,37 @@ So, we can finish our schema like this:
...
+
+
+    class CreatePerson(Mutation):
+        class Arguments:
+            name = String()
+
+        ok = Boolean()
+        person = Field(lambda: Person)
+
+        def mutate(self, info, name):
```
jkimbo:
```suggestion
        def mutate(root, info, name):
```

docs/types/objecttypes.rst
```diff
@@ -62,25 +63,24 @@ passed to the ``ObjectType``.
...
-        last_name = graphene.String()
 
-    class Query(graphene.ObjectType):
-        me = graphene.Field(Person)
-        best_friend = graphene.Field(Person)
+    class Query(ObjectType):
+        me = Field(Person)
+        best_friend = Field(Person)
 
         def resolve_me(_, info):
```
jkimbo:
```suggestion
        def resolve_me(root, info):
```

docs/types/mutations.rst
```diff
@@ -10,14 +10,15 @@ This example defines a Mutation:
...
         class Arguments:
-            name = graphene.String()
+            name = String()
 
-        ok = graphene.Boolean()
-        person = graphene.Field(lambda: Person)
+        ok = Boolean()
+        person = Field(lambda: Person)
 
         def mutate(self, info, name):
```
jkimbo:
```suggestion
        def mutate(root, info, name):
```

docs/types/objecttypes.rst
```diff
@@ -91,10 +91,11 @@ kwargs. For example:
...
-    import graphene
+    from graphene import ObjectType, Field, String
 
-    class Query(graphene.ObjectType):
-        human_by_name = graphene.Field(Human, name=graphene.String(required=True))
+
+    class Query(ObjectType):
+        human_by_name = Field(Human, name=String(required=True))
 
         def resolve_human_by_name(_, info, name):
```
jkimbo:
```suggestion
        def resolve_human_by_name(root, info, name):
```

docs/types/objecttypes.rst
```diff
@@ -170,15 +172,15 @@ A field can use a custom resolver from outside the class:
 
 .. code:: python
 
-    import graphene
+    from graphene import ObjectType, String
 
     def resolve_full_name(person, info):
         return '{} {}'.format(person.first_name, person.last_name)
```
jkimbo:
```suggestion
        return f"{person.first_name} {person.last_name}"
```

docs/relay/nodes.rst
jkimbo: I've not used this way before either but it's quite nice. I guess since we're dropping Python 2 support we should just go all out on Python 3 features.

docs/relay/mutations.rst
jkimbo: I'm all for this. It makes the code much easier to read.

docs/types/objecttypes.rst
jkimbo: Not sure why some of these suggestions have removed a leading space, that wasn't what I was suggesting...

jkimbo:
Oh just saw this is WIP sorry @changeling . Feel free to ignore my comments and I can give it a proper review when the PR is actually ready.

docs/quickstart.rst
jkimbo: I see you're doing this in https://github.com/graphql-python/graphene/pull/984

docs/relay/mutations.rst
changeling: @dvndrsn We had a brief discussion on Slack about this. One reason for this PR as WIP is to further the conversation. See: https://graphenetools.slack.com/archives/CGR4VCHUL/p1559524058000700

@jkimbo I feel the same way.

docs/quickstart.rst
changeling: I went with 3.6+ after looking at what we're testing for in CI. This is absolutely open to discussion!

docs/relay/nodes.rst
changeling: @dvndrsn  I like this for `graphene`, as it feels more pythonic, with the Meta perhaps having it's place in `graphene-django`. 

It's offered in `UPGRADE-v2.0.md` under `Meta as Class arguments`, which may highlight it as best practice to new users/upgrading folks.

Also, `class Meta` seems to offer some confusion among users, if SO is any indication:

https://stackoverflow.com/search?q=%5Bgraphene-python%5D+class+Meta

docs/quickstart.rst
changeling: Absolutely. For this first pass, I (tried!) to ignore actual code outside of the Meta/imports thing, but good catch.

docs/types/mutations.rst
changeling: The no-code done yet, but another good catch.

docs/types/objecttypes.rst
changeling: Ha. No worries. Even if I missed that, I'm trying to blacken the example code, so that would be fixed anyhow, I think.

docs/types/mutations.rst
changeling: This is suggesting the `root` change? Good catch, will fix on second pass.

docs/types/objecttypes.rst
changeling: See previous comments. Got it.

changeling:
> Oh just saw this is WIP sorry @changeling . Feel free to ignore my comments and I can give it a proper review when the PR is actually ready.

Not at all! That's exactly why I made this PR/WIP. To get some more eyes on. Thanks for the feedback and review.

docs/quickstart.rst
changeling: Fixed locally. Will push in a bit.

docs/relay/mutations.rst
changeling: That's a good conversation to have. We need to start a doc/channel to discuss these things. One thought: Any code we use in the docs that mirrors actual code in the examples, IMHO, ought to be taken as is from the example code. What do you think?

changeling: Made a note in my local TODO of this.

examples/starwars_relay/schema.py
changeling: I wanted to preserve this as a reminder to revisit this in conversation. Do we want to go completely explicit, or just top-level to highlight the sub-package. I'm waffling back and forth in my head on this.

examples/complex_example.py
```diff
@@ -1,21 +1,24 @@
+from graphene import InputObjectType, Float, ObjectType, String, Field, Schema
+
+# Naming confusion/conflict with class Mutation and explicit import of graphene.mutation.
```
changeling: I'd love some thoughts on this code. We're importing `Mutation` from `graphene`, previously used as `graphene.Mutation`, _and_ declaring `class Mutation`. 

Side note: We need to do a run through of all the doc examples in a functional context.

docs/types/mutations.rst
dvndrsn: I think the topic of `what is the name of the first argument of a resolver called` is its own docs related discussion/issue. Its definitely not `self`, so this is a good change IMO.

docs/types/mutations.rst
dvndrsn: I made a bunch of changes related to `root` argument in this PR as well: https://github.com/graphql-python/graphene/pull/969

Please nitpick, comment, whatever all you want. That PR and this one are going to conflict in ways that will be annoying to resolve, I'm sure.

examples/complex_example.py
dvndrsn: Yeah, that's no good. This example will need to be improved.

docs/execution/execute.rst
```diff
@@ -40,22 +41,23 @@ You can pass variables to a query via ``variables``.
 
 .. code:: python
 
-    class Query(graphene.ObjectType):
-        user = graphene.Field(User, id=graphene.ID(required=True))
+    class Query(ObjectType):
```
phalt: Whenever I read docs examples like this I always ask "where the hell do I import that from though?".

I feel like if we remove the `graphene` bit we need to show people where it is imported from. Because copy/pasting this won't work.

docs/quickstart.rst
```diff
@@ -35,15 +35,17 @@ one field: ``hello`` and an input name. And when we query it, it should return `
 
 .. code:: python
 
-    import graphene
+    from graphene import ObjectType, Schema, String
```
phalt: This is what I mean about adding imports in docs - we should be doing it consistently.

docs/types/mutations.rst
changeling: Somewhere in the ecosystem, in an `Issue`, there's a great discussion regarding `root` as best, but for the life of me I can't find it again. If I do, I'll pop it into the docs tracking issue.

docs/types/mutations.rst
changeling: Just a thought, but perhaps we ahould look at the two, and consider which will make the most likely to conflict changes and stage them in order to minimize the conflicts? That is, if one makes minor syntactical changes and the other makes bigger changes to actual layout and wording...?

docs/execution/execute.rst
changeling: This is something I've been meaning to bring up. I think we should, wherever possible, make snippets in the docs complete. That is, again as much as possible, allow them to be copy/pasted into a `.py` file and actually work. I echo your thought here.

docs/quickstart.rst
changeling: Agreed. See above.

changeling: Is there a sphinx or other tool for actually embedding live code from, say, `examples/`, for inclusion inline in the docs?

docs/execution/execute.rst
jkimbo: +1 on including imports. I always get confused with that too.

maquino1985:
Comment on the use of graphene. graphene. graphene. everywhere. At first I thought this was annoying and didn’t understand why it was used in the docs but I’ve got back to importing graphene types this way because they have the same name as typing types. 

On that note, is there a reason that this library doesn’t use the typing module? It would at least help new users like myself to figure out how to get things to work until this project starts moving more quickly on its various issues.

docs/quickstart.rst
dvndrsn: This might be what you're looking for @changeling `literalinclude`:

https://quick-sphinx-tutorial.readthedocs.io/en/latest/code.html

This option `emphasize-lines` is interesting too.

docs/relay/mutations.rst
dvndrsn: I really like this idea to extract most examples from a working API. I had a few ideas that I brainstormed in this issue:

https://github.com/graphql-python/graphene/issues/970

docs/relay/nodes.rst
dvndrsn: Yeah, that's true. Especially with Graphene-Django and Graphene-SQLAlchemy. Both of those projects have a lot of use of `Meta` parameters.

docs/types/enums.rst
```diff
@@ -32,17 +33,16 @@ needs to have the ``description`` property on it.
 
 .. code:: python
 
-    class Episode(graphene.Enum):
+    class Episode(Enum):
```
dvndrsn: Definitely need an import in this example for disambiguation.

```suggestion
    from graphene import Enum
    
    class Episode(Enum):
```

docs/types/mutations.rst
dvndrsn: I think we we can resolve the conflicts easily enough, but maybe if we can groom some stories around documentation there'll be less overlap.

jkimbo:
@changeling were you planning on updating this PR?

borisowsky:
@changeling what state on 3.0?

docs/execution/execute.rst
radekwlsk: Also importing a package would be a better approach rather than import each class separately. Makes reading code much easier if each external reference is easily identifiable.

docs/relay/nodes.rst
radekwlsk: I feel like docs should at least present both approaches. As I tend to use `Meta` anyway in my Django project, where `graphene-django` is also used, it will be nice to have a clear indication that you can do that with `graphene` also. That may not be clear for folks less familiar with Python.

docs/types/enums.rst
```diff
@@ -52,16 +52,16 @@ the ``Enum.from_enum`` function.
 
 .. code:: python
 
-    graphene.Enum.from_enum(AlreadyExistingPyEnum)
+    Enum.from_enum(AlreadyExistingPyEnum)
```
radekwlsk, line 55: Like mentioned in https://github.com/graphql-python/graphene/pull/992/files#r291840336 each use of class that also exists in standard lib should have import. Best if all examples have all imports required. I personally would keep `import graphene` and `graphene.xxx` as that is a good practice to import packages rather than members.

docs/types/mutations.rst
radekwlsk: I think that every method where `root` or `parent` is used as the first argument it should be decorated with `@staticmethod`. Especially if that argument is not an instance of the class. Current approach is not too pythonic.

examples/complex_example.py
radekwlsk: And that is exactly why you should just `import graphene` so each external dependency is always clearly distinguishable as external.

PR #995, opened 2019-06-05
Fix malformed version on aniso8601 requirement

AdrianLC:
This was changed on https://github.com/graphql-python/graphene/pull/935 to fix a warning when installing newer versions but we are now getting an error that 6.0.0 doesn't match 6.0.*:
 ```
ERROR: graphene 2.1.5 has requirement aniso8601<=6.0.*,>=3, but you'll have aniso8601 6.0.0 which is incompatible.
```
I don't think that the `*` are recognized as the same version.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -87,7 +87,7 @@ setup(
         "six>=1.10.0,<2",
         "graphql-core>=2.1,<3",
         "graphql-relay>=0.4.5,<1",
-        "aniso8601>=3,<=6.0",
+        "aniso8601>=3,<=6",
     ],
     tests_require=tests_require,
     extras_require={
--- a/setup.py
+++ b/setup.py
@@ -87,7 +87,7 @@ setup(
         "six>=1.10.0,<2",
         "graphql-core>=2.1,<3",
         "graphql-relay>=0.4.5,<1",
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: setup.py [hunk 1]
```diff
@@ -87,7 +87,7 @@ def run_tests(self):
         "six>=1.10.0,<2",
         "graphql-core>=2.1,<3",
         "graphql-relay>=0.4.5,<1",
-        "aniso8601>=3,<=6.0.*",
+        "aniso8601>=3,<=6.0",
```
jkimbo:
```suggestion
        "aniso8601>=3,<=6",
```
What do you think @AdrianLC ? That way it will accept minor versions too?

AdrianLC: Even better IMO, I only left it as 6.0 to not change the behavior intended on #935

benhowes:
For those who are using pipenv, this problem does also seem to be a cause of:
```
ERROR: Could not find a version that matches aniso8601<8.0.0,<=6.0.*,>=3,>=7.0.0
```
When locking. In order to get my locking stable, I had to switch out the released version for `pipenv install -e https://github.com/graphql-python/graphene.git` for now.

PR #1002, opened 2019-06-09
Remove sponsorship text from readme

mvanlonden:
As discussed in the last meetup, this removes the sponsorship text from the README.


ROADMAP.md
```diff
@@ -17,7 +17,6 @@ The goal is to improve adoption and sustainability of the project.
   - Create [Patreon page](https://www.patreon.com/syrusakbary)
```
changeling: Are we keeping a Patreon page on the ROADMAP?

mvanlonden: Good catch. Updated accordingly

mvanlonden:
Also removed from the graphene-python landing page https://github.com/graphql-python/graphene-python.org/pull/3

PR #1003, opened 2019-06-09
[WIP] Build docs with vanilla read the docs template

mvanlonden:
Uses `sphinx_rtd_theme` with more functionality like version toggle UI.


changeling:
Let's hold this until after we have the dual docs setup, to see what that looks like

PR #1007, opened 2019-06-12
Don't delete the Meta class

jkimbo:
To support use cases like https://github.com/graphql-python/graphene-django/issues/671 where analysis is done on attributes defined on the Meta class.

I'm not very familiar with this code so I'm not sure if there are any implications of this change but all the tests are passing so 🤷‍♀

### DIFF
--- a/graphene/utils/subclass_with_meta.py
+++ b/graphene/utils/subclass_with_meta.py
@@ -37,7 +37,6 @@ class SubclassWithMeta(six.with_metaclass(SubclassWithMeta_Meta)):
                         _Meta
                     )
                 )
-            delattr(cls, "Meta")
         options = dict(meta_options, **_meta_props)
 
         abstract = options.pop("abstract", False)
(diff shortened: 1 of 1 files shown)


graphene/utils/subclass_with_meta.py
```diff
@@ -46,10 +45,10 @@ def __init_subclass__(cls, **meta_options):
...
                 "Received: abstract, {option_keys}"
             ).format(option_keys=", ".join(options.keys()))
-        else:
-            super_class = super(cls, cls)
-            if hasattr(super_class, "__init_subclass_with_meta__"):
-                super_class.__init_subclass_with_meta__(**options)
+
+        super_class = super(cls, cls)
+        if hasattr(super_class, "__init_subclass_with_meta__"):
+            super_class.__init_subclass_with_meta__(**options)
```
jkimbo: Without this change `__init_subclass_with_meta__` wouldn't be called on any subclass of a class that was defined as abstract.

syrusakbary:
Closing the issue as the information provided by `Meta` should be available in `_meta` (the issue that the person linked was more related to renaming a field from `only_fields` to `fields`)

jkimbo:
@syrusakbary that is not quite right, only the fields that are selectively selected from the Meta class are stored on the _meta instance which limits its usefulness a bit. This is probably the wrong way to go about this though so I'll open another PR.

PR #1033, opened 2019-07-15
Update required versions of core and relay

Cito:
Graphene should require graphql-relay-py>=2
because older graphql-relay-py doesn't work with graphql-core 2.


Cito:
This is a duplicate of #1032, so closing this.

PR #1074, opened 2019-09-26
Add a helpful message to when a global_id fails to parse.

allen-munsch:
It was really difficult to determine that I wasn't using the correct format for an ID.

It was previously returning null, it now returns an error.message with `Incorrect Padding` and displays the exception traceback inside the server logs.

I'm not sure about whether `return None` is/was intended functionality.

I do see some of the tests are failing:

```
_________________________ test_node_query_incorrect_id _________________________
    def test_node_query_incorrect_id():
        executed = schema.execute(
            '{ node(id:"%s") { ... on MyNode { name } } }' % "something:2"
        )
>       assert not executed.errors

E       assert not [GraphQLError("Unable call from_global_id, is the id base64 encoding of 'TypeName:id': something:2 Exception: Incorrect padding", 
locations=[SourceLocation(line=1, column=3)], path=['node'])]

E        +  where [GraphQLError("Unable call from_global_id, is the id base64 encoding of 'TypeName:id': something:2 Exception: Incorrect padding", 
locations=[SourceLocation(line=1, column=3)], path=['node'])] = 
ExecutionResult(data={'node': None}, 
errors=[GraphQLError("Unable call from_global_id, is the id base64 encoding of 'TypeName:id': something:2 Exception: Incorrect padding", 
locations=[SourceLocation(line=1, column=3)], 
path=['node'])]).errors
```

edit: test updated.

### DIFF
--- a/graphene/relay/node.py
+++ b/graphene/relay/node.py
@@ -1,3 +1,4 @@
+from collections import OrderedDict
 from functools import partial
 from inspect import isclass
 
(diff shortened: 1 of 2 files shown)


kazamatzuri:
The tests are not quite passing yet: 

https://travis-ci.org/graphql-python/graphene/jobs/624235847?utm_medium=notification&utm_source=github_status

jkimbo:
@allen-munsch thanks for the contribution. I think this is a great addition and I've updated your PR and expanded the checks to cover more error cases.

PR #1080, opened 2019-10-04
Use unidecode to handle unicode characters in constant names

### LINKED ISSUE #1079: to_const doesn't support unicode
Let's say I have a model in Django with a field that has choices. Only, because I'm Icelandic, those choices might be in Icelandic and contain utf-8 characters. Generating an enum from my model will fail.

F.ex.:
```python
class School(models.Model):
    MUNICIPALITIES = (
        ('Hafnarfjarðarkaupstaður', 'Hafnarfjarðarkaupstaður'),
        ('Reykjanesbær', 'Reykjanesbær'),
        ('Garðabær', 'Garðabær'),
        ...
    )
    name = models.CharField(max_length=128)
    municipality = models.CharField(max_length=32, blank=True, null=True, choices=MUNICIPALITIES)
```

I then try to create a Query object:
```python
class SchoolNode(DjangoObjectType):
    @classmethod
    class Meta:
        model = School
        filter_fields = [
            'id',
            'name',
        ]
        exclude_fields = [
        ]
        interfaces = (relay.Node,)


class Query(object):
    school = relay.Node.Field(SchoolNode)
    all_schools = DjangoFilterConnectionField(SchoolNode)
```
This will then fail in `graphql.utils.assert_valid_name.assert_valid_name` at schema evaluation since to_const doesn't handle converting unicode characters to ascii:
`Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but "A_HAFNARFJARÐARKAUPSTAÐUR" does not.`

I really doubt I'm the only one running into this. Most people just handle this stuff differently I suppose? Manually switch out code, or use an actual Enum() in the choices?

My proposed solution is to use the unidecode package, specifically designed for this. I'll throw together a PR.

henrythor:
This fixes #1079

### DIFF
--- a/graphene/utils/str_converters.py
+++ b/graphene/utils/str_converters.py
@@ -1,4 +1,5 @@
 import re
+from unidecode import unidecode
 
 
 # Adapted from this response in Stackoverflow
@@ -18,4 +19,4 @@ def to_snake_case(name):
 
 
 def to_const(string):
-    return re.sub(r"[\W|^]+", "_", string).upper()  # noqa
+    return re.sub(r"[\W|^]+", "_", unidecode(string)).upper()  # noqa
--- a/graphene/utils/str_converters.py
+++ b/graphene/utils/str_converters.py
@@ -19,4 +19,4 @@ def to_snake_case(name):
 
 
 def to_const(string):
-    return re.sub(r"[\W|^]+", "_", unidecode(string)).upper()  # noqa
+    return re.sub(r"[\W|^]+", "_", unidecode(string)).upper()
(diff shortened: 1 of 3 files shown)


henrythor:
I have no idea why the snapshots are failing in pytest, seems unlikely they're affected by my patch :/

jkimbo:
This is still an issue.

jkimbo:
@henrythor thanks the PR. I've rebased your changes on top of master so that the tests pass (you were right the tests were failing for unrelated reasons).

Issue #1087, opened 2019-10-29
Example in types/objecttypes is out of date


tompao:
1. namedtuple is not created correctly
2. result object is not subscriptable as result['data']
3. result data key is incorrect (my_best_friend -> myBestFriend)

PR #1088, opened 2019-10-29
Fix objecttypes DefaultResolver example (#1087)

tompao:
* Create namedtuple as expected
* Access result.data instead of result['data']
* Refer to field with camel-case name


jkimbo:
Thanks @tompao !

PR #1111, opened 2019-12-11
The default_value of InputField should be INVALID

ganwell:
Since GraphQL 3.0 there is a distinction between None and INVALID (no value).
The tests captured the bug and are updated.

See also https://github.com/graphql-python/graphql-core-next/pull/66#issuecomment-557688621

I think there are more bugs of this kind, but since I don't know graphene well, yet, I take baby steps.

I could not track the snapshot differences, the order is wrong and master has the same problem too. If master is solved I'll rebase this PR. @jkimbo has the same differences in his commits.


jkimbo:
@ganwell I've updated your PR to use `Undefined` instead of `INVALID` to keep up to date with graphql-core: https://github.com/graphql-python/graphql-core-next/issues/77

PR #1139, opened 2020-02-21
Fix example query in quickstart doc

ko-lem:
From the "Querying" section of the quickstart doc:
```
# we can query for our field (with the default argument)
query_string = '{ hello }'
result = schema.execute(query_string)
print(result.data['hello'])
# "Hello stranger"

# or passing the argument in the query
query_with_argument = '{ hello(name: "GraphQL") }'
result = schema.execute(query_with_argument)
print(result.data['hello'])
# "Hello GraphQL!"
```
(https://docs.graphene-python.org/en/latest/quickstart/#querying)

"Hello stranger" should also end with a `!`


jkimbo:
Thanks @ko-lem !

PR #1143, opened 2020-02-26
Fixed import causing Graphene v3 to crash

### LINKED ISSUE #1142: Graphene v3 crashes on import
I'm running some benchmarks on different Python GraphQL servers, and tried testing with the dev version of Graphene v3 from PyPi. When importing the library, I got the following error:

```python
  File "./server.py", line 4, in <module>
    import graphene
  File "venv/lib/python3.7/site-packages/graphene/__init__.py", line 3, in <module>
    from .types import (
  File "venv/lib/python3.7/site-packages/graphene/types/__init__.py", line 8, in <module>
    from .datetime import Date, DateTime, Time
  File "venv/lib/python3.7/site-packages/graphene/types/datetime.py", line 6, in <module>
    from graphql.error import INVALID
ImportError: cannot import name 'INVALID' from 'graphql.error' (venv/lib/python3.7/site-packages/graphql/error/__init__.py)
```

It looks like `graphene/types/datetime.py` tried to import `INVALID` from the wrong module (`graphql.error` in instead of `graphql`).

jaydenwindle:
This PR fixes #1142 by importing `INVALID` from `graphql` instead of `graphql.errors` in `graphene/types/datetime.py`.

### DIFF
--- a/graphene/types/datetime.py
+++ b/graphene/types/datetime.py
@@ -3,7 +3,7 @@ from __future__ import absolute_import
 import datetime
 
 from aniso8601 import parse_date, parse_datetime, parse_time
-from graphql.error import INVALID
+from graphql import INVALID
 from graphql.language import StringValueNode
 
 from .scalars import Scalar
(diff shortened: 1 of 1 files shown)


Cito:
Thanks for the fix. Since you're at it already, you should also change this to use `Undefined` instead of  `INVALID` which has been deprecated.

jaydenwindle:
@Cito Good call, will do :)

PR #1153, opened 2020-03-14
Improve enum compatibility

jkimbo:
This PR improves enum compatibility by allowing resolvers to return enums as well as their values and names. It also changes the input values for mutations to be the enum rather than the enum value.

Example:

```python
from enum import Enum as PyEnum

class Color(PyEnum):
    RED = 1
    GREEN = 2
    BLUE = 3

GColor = Enum.from_enum(Color)

class Query(ObjectType):
    color = GColor(required=True)
    color_by_value = GColor(required=True)

    def resolve_color(_, info):
        return Color.RED

    def resolve_color_by_value(_, info):
        return Color.RED.value

schema = Schema(query=Query)

results = schema.execute(
    """query {
      color
      colorByValue
    }"""
)
assert not results.errors

assert results.data["color"] == "RED"
assert results.data["colorByValue"] == "RED"
```

Since this change is backwards incompatible by changing the way mutations with enum inputs work I've also added a compatibility middleware (`enum_value_convertor_middleware`) that can be added so that mutations behave as they did. This should give people the ability to migrate their applications after upgrading to v3.

## TODO

* [x] Convert examples and other tests (like in this PR: https://github.com/graphql-python/graphene/pull/879/files)
* [x] Add some tests for the compat middleware
* [ ] Create upgrade to v3 guide


syrusakbary:
I think this PR needs further review and deep analysis. I'm not sure the middleware approach is the optimal one to solve the compatibility with Graphene 2

syrusakbary:
Looks great. Merging

fescobar:
Hi @jkimbo 
I would like to know when it will be released version 3 of `graphene-django`?
https://pypi.org/project/graphene-django/#history
There is no updates for a long time ago.
Thank you in advance.

PR #1157, opened 2020-03-14
Allow fast ObjectType creation based on dataclasses

syrusakbary:
This PR optimizes the ObjectType initialization by an order of magnitude by leveraging the same strategy that dataclasses introduced (dynamic creation of optimal `__init__` functions based on eval).

## Benchmarks

This are the benchmarking numbers after running:
```
py.test --benchmark-only graphene/types/tests/test_objecttype.py
```

Before this patch (385,925 operations / second)

```
-------------------------------------------------------- benchmark: 1 tests -------------------------------------------------------
Name (time in us)                          Min      Max    Mean  StdDev  Median     IQR  Outliers  OPS (Kops/s)  Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------------------------
test_objecttype_container_benchmark     1.7730  61.1060  2.5912  1.3161  2.4700  0.2210  911;2832      385.9257   46230           1
-----------------------------------------------------------------------------------------------------------------------------------
```

After this patch (1,270,600 operations / second) - **More than 3x speedup**

```
---------------------------------------------------------------- benchmark: 1 tests ----------------------------------------------------------------
Name (time in ns)                            Min           Max      Mean      StdDev    Median       IQR  Outliers  OPS (Mops/s)  Rounds  Iterations
----------------------------------------------------------------------------------------------------------------------------------------------------
test_objecttype_container_benchmark     594.9987  103,029.9973  787.0462  1,168.2685  708.9984  144.9953  752;3399        1.2706  119389           1
----------------------------------------------------------------------------------------------------------------------------------------------------
```

## Breaking change

This PR introduces a breaking change, private attributes (aka: `_*` fields) would not be automatically modified on instantiation.

```python
class MyObject(ObjectType):
    field = String()

MyObject(_private_field=1)  # <--- This will not work, unless you create a custom __init__
```

Note: DO NOT MERGE YET


graphene/pyutils/dataclasses.py
```diff
@@ -0,0 +1,1260 @@
+# This is a polyfill for dataclasses
```
jkimbo: Why not add a dependency on the `dataclasses` lib?

graphene/types/objecttype.py
```diff
@@ -14,7 +19,34 @@ class ObjectTypeOptions(BaseOptions):
...
+        # We create this type, to then overload it with the dataclass attrs
+        class InterObjectType:
+            pass
+
+        base_cls = super().__new__(cls, name, (InterObjectType,) + bases, namespace)
+        if base_cls._meta:
+            fields = [
+                (
+                    key,
+                    "typing.Any",
```
jkimbo: Can this typing be more accurate by taking the type of the GraphQL field?

jkimbo:
Also the breaking change could be quite a big one. Can you give an example of a custom `__init__` method that would let you define private attributes?

graphene/pyutils/dataclasses.py
syrusakbary: dataclasses are not available since Python 3.7

graphene/types/objecttype.py
syrusakbary: Yes, that would be ideal. However I have a bigger plan that we can action in a separate PR.

syrusakbary:
> Also the breaking change could be quite a big one. Can you give an example of a custom __init__ method that would let you define private attributes?

You can overwrite your `__init__` allowing the extra private fields. There is a hacky solution we can get to skip the breaking change, but I think a better long-term solution is just removing it's support and focusing on documentation.

graphene/pyutils/dataclasses.py
jkimbo: No I mean the dataclasses library on PyPi: https://pypi.org/project/dataclasses/

syrusakbary: We might want to customize it a bit, so I think is better to ship it with the lib for now and remove it later :)

syrusakbary:
After thinking on it a bit more, I think we can test the water by merging the PR and seeing if anyone has significant issues over the breaking change. So we should be good to merge! YAY!

wodCZ:
Hi,
just a thought - could this change be backported to v2? From a quick glance it looks like the changed code is almost identical to what's in v2. 
I'm not sure whether there are some other changes elsewhere on which this depends on.

Thanks!

PR #1170, opened 2020-04-01
Add note about the use of `args`

### LINKED ISSUE #1037: Field argument names conflict with constructor arguments
The [current implementation](https://github.com/graphql-python/graphene/blob/167c8c203cc2607fd3b80359de4b6064ac2e5234/graphene/types/field.py#L63) of `Field`'s constructor handles arguments named `name` or `source` that are of `Argument` type but does not do the same for the rest of them (e.g. `description` or `required`).

For example, this fails as follows:
```python
import graphene

    answer = graphene.Int(description=graphene.Int())
    def resolve_answer(self, info, description):
        return description

result = schema.execute('{ answer(description: 42) }')

assert result.data
assert result.data['answer'] == 42
```

jkimbo:
Closes #1037

### DIFF
--- a/graphene/types/field.py
+++ b/graphene/types/field.py
@@ -40,11 +40,12 @@ class Field(MountedType):
             last_name = graphene.Field(String, description='Surname')  # explicitly mounted as Field
 
     args:
-        type (class for a graphene.UnmountedType): must be a class (not an instance) of an
+        type (class for a graphene.UnmountedType): Must be a class (not an instance) of an
             unmounted graphene type (ex. scalar or object) which is used for the type of this
             field in the GraphQL schema.
-        args (optional, Dict[str, graphene.Argument]): arguments that can be input to the field.
-            Prefer to use **extra_args.
+        args (optional, Dict[str, graphene.Argument]): Arguments that can be input to the field.
+            Prefer to use ``**extra_args``, unless you use an argument name that clashes with one
+            of the Field arguments presented here (see :ref:`example<Scalars>`).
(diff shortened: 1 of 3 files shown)


tdiam:
Some proposed changes [here](https://github.com/graphql-python/graphene/compare/update-docs-field-args...tdiam:update-docs-field-args).

jkimbo:
Thanks @tdiam . Updated with your suggestions

graphene/types/field.py
```diff
@@ -40,11 +40,12 @@ class Person(ObjectType):
...
     args:
-        type (class for a graphene.UnmountedType): must be a class (not an instance) of an
+        type (class for a graphene.UnmountedType): Must be a class (not an instance) of an
             unmounted graphene type (ex. scalar or object) which is used for the type of this
             field in the GraphQL schema.
-        args (optional, Dict[str, graphene.Argument]): arguments that can be input to the field.
-            Prefer to use **extra_args.
+        args (optional, Dict[str, graphene.Argument]): Arguments that can be input to the field.
+            Prefer to use ``**extra_args``, unless you use an argument name that clashes with one
+            of the Field arguments presented here (see :ref:`example<Scalars>`).
```
tdiam, line 48: `Scalars`: Don't you mean `ObjectType` where the example for using `args` is?

jkimbo: Hmm yeah I was actually thinking the Scalars page had a nice list of the reserved but I think you're right: it makes sense to point back to the ObjectType page.

PR #1173, opened 2020-04-03
Section added `GraphQL testing clients`

Nishchit14:


syrusakbary:
While I think is useful to add docs about different testing clients, I don't think the main README is the place for it. At the same time, I had a different thing in mind when thinking about testing clients.

Perhaps I think this PR will be better in a project exposing different visual GraphQL clients (instead of than in backend projects like Graphene).

Also, great work on Firecamp, I didn't know it existed until today!

Nishchit14:
Thank you for reviewing the PR @syrusakbary, I completely understand your view here.

PR #1174, opened 2020-04-04
Update __init__.py

rrwt:
Edited to add datetime to init. This will help us with module discovery


rrwt:
@jkimbo Thank you. I didn't see that it has been fixed for the latest version. Just checked and my org is using a different version.

PR #1176, opened 2020-04-08
Fixed code error in document

asvrada:
Changed JavaScript syntax to Python one.

### DIFF
--- a/docs/relay/mutations.rst
+++ b/docs/relay/mutations.rst
@@ -22,8 +22,8 @@ subclass of ``relay.ClientIDMutation``.
 
         @classmethod
         def mutate_and_get_payload(cls, root, info, **input):
-            ship_name = input.ship_name
-            faction_id = input.faction_id
+            ship_name = input["ship_name"]
+            faction_id = input["faction_id"]
             ship = create_ship(ship_name, faction_id)
             faction = get_faction(faction_id)
             return IntroduceShip(ship=ship, faction=faction)


syrusakbary:
I'm not sure if I understand. Using object attributes is also Pythonic, in fact we did the addition that resolver arguments are inside a python object (instead of a dict) to make it much easier to reason about.

I'm curious to hear thoughts from more people here

asvrada:
Using object attributes won't work in my project, might be I did something wrong, could you have a look? Thanks!

```python
# schema.py
class CreateIcon(relay.ClientIDMutation):
    class Input:
        keyword = graphene.String(required=True)
        path = graphene.String(required=True)

    # output
    icon = graphene.Field(IconType)

    @classmethod
    def mutate_and_get_payload(cls, root, info, **payload):
        keyword = payload.keyword
        path = payload["path"]
        icon = models.Icon.objects.create(keyword=keyword, path=path)
        return CreateIcon(icon=icon)

class Mutation(graphene.ObjectType):
    create_icon = CreateIcon.Field()
```
Would throw error:

```
Bad Request: /graphql/
[13/Apr/2020 xx:48:18] "POST /graphql/ HTTP/1.1" 400 116
Traceback (most recent call last):
  File "yyy/python3.7/site-packages/promise/promise.py", line 489, in _resolve_from_executor
    executor(resolve, reject)
  File "yyy/python3.7/site-packages/promise/promise.py", line 756, in executor
    return resolve(f(*args, **kwargs))
  File "yyy/python3.7/site-packages/graphql/execution/middleware.py", line 75, in make_it_promise
    return next(*args, **kwargs)
  File "yyy/python3.7/site-packages/graphene/relay/mutation.py", line 70, in mutate
    result = cls.mutate_and_get_payload(root, info, **input)
  File "xxx/bill/schema.py", line 83, in mutate_and_get_payload
    keyword = payload.keyword
graphql.error.located_error.GraphQLLocatedError: 'dict' object has no attribute 'keyword'

```

hf-krechan:
I have the same error.
It would be nice to use the attribute notations. So the IDE and linter-tools could support you.
Maybe one could make use of the `namedtuple` to create an object from a dictionary behind the scences:
```python
from collections import namedtuple

input = {'first_name': 'Arty', 'last_name': 'Fischel'}
input_object = namedtuple('input_object', input.keys())(*input.values())
print(input_object.first_name)
```

But for now I would stick to the `input.get("first_name")` approach to avoid errors.

Issue #1180, opened 2020-04-10
Unable to compare two instances of graphene.ObjectType classes


spacether:
`__eq__` is not defined on graphene.ObjectType classes
Please add it so we can compare two instances of classes which inherit from graphene.ObjectType
using `self.assertEquals(obj, obj_other)`

KingDarBoja:
This issue can be closed as solved at #1157

PR #1181, opened 2020-04-12
Adds __eq__ and test of it

spacether:
This PR adds the `__eq__` method to the ObjectType class so instances of it can be compared
Tests of the new method has also been added.
Happy to add more tests my guess is that we may need to cover more use cases.

If merged, this will close out [this issue](https://github.com/graphql-python/graphene/issues/1180)

### DIFF
--- a/graphene/types/objecttype.py
+++ b/graphene/types/objecttype.py
@@ -118,6 +118,11 @@ class ObjectType(BaseType, metaclass=ObjectTypeMeta):
         **kwargs (Dict[str: Any]): Keyword arguments to use for Field values of value object
     """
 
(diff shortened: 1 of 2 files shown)


syrusakbary:
We just merged #1157, which indirectly added support for `__eq__` in Objects.

https://github.com/graphql-python/graphene/blob/master/graphene/types/tests/test_objecttype.py#L167-L173

Could test without the `__eq__` method addition but without any change on the ObjectType logic itself? (just adding the test if seeing if that passes)

spacether:
This PR is no longer needed because https://github.com/graphql-python/graphene/pull/1157 added the __eq__ method

Issue #1110, opened 2019-12-10
examples are added to site-packages


marnixlenoble:
When installing graphene using pip it creates an examples directory in site-packages with the star wars example schema.

I feel like it's improper to add a base examples directory. Either it should not be added or placed inside the graphene directory.

Steps to reproduce:

Start empty pipenv project

Run `pipenv install graphene`

In a python file: `from examples.starwars.schema import schema`

radekwlsk:
@marnixlenoble @jkimbo I've found that issue cause in `setup.py` and I could fix it.  Unfortunately there seems to be no easy way to make package `examples` an optional dependency when it is in the same repository. 
Maybe it will be a good idea to separate examples from main package and then it could be added as `extras_require={"examples": ["graphene-examples"]}`? Is it even necessary to allow installation of examples or are they here only for documentation purposes?

Mc01:
Seems like good starting point for `cleanup package content and setup.py` PR 😉 
@radekwlsk maybe you should provide solution proposal as new pull request with label `enhancement`?

radekwlsk:
@jkimbo #1187 resolved this issue, you may link the two together and close it.

jkimbo:
Yep this can be closed now.

PR #1187, opened 2020-04-17
Update excluded packages list to properly exclude examples package

radekwlsk:
As #1110 correctly states, currently `examples` package is installed alongside `graphene` which is probably not intended as `"examples"` is listed in `packages=find_packages(exclude=[...])` list.

I was able to fix it by changing examples in:
`packages=find_packages(exclude=["tests", "tests.*", "examples"])`
 to examples*:
`packages=find_packages(exclude=["tests", "tests.*", "examples*"])`

Unfortunately there seems to be no easy way to make package examples an optional dependency when it is in the same repository. Maybe it will be a good idea to separate examples from main package and then it could be added as extras_require={"examples": ["graphene-examples"]}?

Is it even necessary to allow installation of examples or are they here only for documentation purposes?

_Additional question:_
I am also not sure what `"tests"` in exclude is supposed to do as all test files are currently also included in package installation.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -80,7 +80,7 @@ setup(
         "Programming Language :: Python :: 3.8",
     ],
     keywords="api graphql protocol rest relay graphene",
-    packages=find_packages(exclude=["tests", "tests.*", "examples"]),
+    packages=find_packages(exclude=["tests", "tests.*", "examples*"]),
     install_requires=[
         "graphql-core>=3.1.0b1,<4",
         "graphql-relay>=3.0,<4",


radekwlsk:
@jkimbo Thanks for merging it so fast! Glad to make my first OSS contribution 🔥 

I will try to get on tests soon, it is possible that adding `*` to `tests` in excluded list will also do the trick.

radekwlsk:
@jkimbo After further investigating the `tests` packages in installation I've found that installing Graphene from source with 
```bash
$ pip install .
``` 
or from PyPI with 
```bash
$ pip install graphene
``` 
both include `tests`. 
But when first building a source distribution and the installing from it with 
```bash
$ python setup.py sdist
$ pip install dist/graphene-3.0b1.tar.gz
```
then the `tests` packages are indeed excluded even with current setup config.

Maybe the problem lies in how the Graphene is distributed through PyPI, not with the config?

jkimbo:
That is possible @radekwlsk . Also I'm not sure that excluding tests is a good idea actually because there are some cases where you would want to run the tests from the distribution. See https://github.com/graphql-python/graphene-django/issues/820

Do you have any objections to keeping the tests included as part of the release?

radekwlsk:
> Do you have any objections to keeping the tests included as part of the release?

None at all, I've just figured that was the intention behind `find_packages(exclude=["tests", "tests.*", ...])`. I am well ok with that.

jkimbo:
Yep I think that was the original intention but it seems that it's useful for some cases. I'll remove it from the exclude list.

PR #1219, opened 2020-06-27
ObjectType meta arguments

jkimbo:
This PR updates ObjectType to allow arguments to be passed to the metaclass.

### DIFF
--- a/graphene/types/objecttype.py
+++ b/graphene/types/objecttype.py
@@ -20,13 +20,32 @@ class ObjectTypeOptions(BaseOptions):
 
 
 class ObjectTypeMeta(BaseTypeMeta):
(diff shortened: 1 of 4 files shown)


graphene/types/objecttype.py
```diff
@@ -20,12 +20,14 @@ class ObjectTypeOptions(BaseOptions):
 
 
 class ObjectTypeMeta(BaseTypeMeta):
-    def __new__(cls, name, bases, namespace):
+    def __new__(cls, name_, bases, namespace, **kwargs):
```
syrusakbary, line 23: We shouldn't use `**kwargs` here. It's intended that the `ObjectTypeMeta` (and any other `*Meta`) have explicit defined values and types.

PR #1221, opened 2020-07-01
Add Base64 scalar

EpicEric:


graphene/types/base64.py
```diff
@@ -0,0 +1,45 @@
+from __future__ import absolute_import
```
jkimbo, line 1: I don't think you need this import

graphene/types/tests/test_base64.py
```diff
@@ -0,0 +1,101 @@
...
+        return 42
+
+
+schema = Schema(query=Query)
+
+
+def test_base64_query():
+    base64_value = base64.b64encode(b"Random string").decode("utf-8")
+    result = schema.execute(
+        """{ base64(input: "%s", match: "Random string") }""" % base64_value
```
jkimbo, line 38:
```suggestion
        """{ base64(input: "{}", match: "Random string") }""".format(base64_value)
```

EpicEric: Using `{}` to format this query means escaping other curly brackets -- which is probably the reason the old formatting is used --, but it can be done:

```diff
-        """{ base64(input: "%s", match: "Random string") }""" % base64_value
+        """{{ base64(input: "{}", match: "Random string") }}""".format(base64_value)
```

I've applied the change above.

jkimbo: Oh good point. Thanks for making the change

Issue #729, opened 2018-05-20
A new way to write ObjectType with python3's annotation


ocavue:
_This issue is a **feature discussion**._

For now, the way to write a ObjectType is as below:

```Python
class Hero(ObjectType):
    name = Field(List(String), only_first_name=Boolean())
    age = Int()

    def resolve_name(self, info, only_first_name=False):
        if only_first_name:
            return [self._first_name]
        return [self._first_name, self._last_name]

    def resolve_age(self, info):
        return self._age
```

I define name twice (`Hero.name` and `Hero.resolve_name`) because I have to define types of arguments and return value. This will cause some reading and writing problems.

Since python3, annotation feature bring a native way to describe types. By using annotation, we can rewrite this class with a clearer way:

```python
class Heroine(AnnotationObjectType):

    def name(self, info, only_first_name: Boolean() = False) -> List(String):
        if only_first_name:
            return [self._first_name]
        return [self._first_name, self._last_name]

    def age(self, info) -> Int():
        return self._age


print(Heroine.name.__annotations__)
# {'only_first_name': <graphene.types.scalars.Boolean object at 0x104cb8550>, 'return': <graphene.types.structures.List object at 0x105742f28>}
```

`AnnotationObjectType` shouldn't be difficult to write if we somehow transform it into `ObjectType`. But I think a looooot of cases should be tested before being released.

Of cause even if we write an `AnnotationObjectType` class, `ObjectType` class will not be dropped since python2.7 doesn't support annotation.

I would like to hear you guys' comments and suggestions before doing this.

ekampf:
Very interesting!

jkimbo:
I really like this @ocavue ! I had no idea you could do this python annotations. I don't know much about annotations actually but I have a couple of questions:

* How would ObjectTypes work?
* Could you use both ObjectTypes and AnnotatedObjectType's in the same schema?
* Is there anything that can be expressed in the python type system that can't be expressed in GraphQL (or vice versa)?

I think keeping the ObjectType and AnnotatedObjectType separate is essential, not just for keeping python 2.7 compatibility, but also supporting both ways of writing ObjectTypes.

ocavue:
@jkimbo here are my answers. I'm not very familiar with Graphene nor Annotation. So any corrections are welcome.

> How would ObjectTypes work?

[...]

PR #1224, opened 2020-07-03
Allow adding GraphQL types to Graphene schema

jkimbo:
This PR allows adding plain GraphQL types to a Graphene schema. This opens up the option for other libraries to interop with Graphene.

### DIFF
--- a/graphene/types/objecttype.py
+++ b/graphene/types/objecttype.py
@@ -1,3 +1,5 @@
+import inspect
+
 from .base import BaseOptions, BaseType, BaseTypeMeta
 from .field import Field
 from .interface import Interface
@@ -137,7 +139,7 @@ class ObjectType(BaseType, metaclass=ObjectTypeMeta):
         fields = {}
 
         for interface in interfaces:
-            assert issubclass(
+            assert inspect.isclass(interface) and issubclass(
                 interface, Interface
             ), f'All interfaces of {cls.__name__} must be a subclass of Interface. Received "{interface}".'
             fields.update(interface._meta.fields)
--- a/graphene/types/schema.py
+++ b/graphene/types/schema.py
@@ -23,6 +23,7 @@ from graphql import (
     GraphQLObjectType,
     GraphQLSchema,
     GraphQLString,
+    GraphQLType,
     Undefined,
 )
 
@@ -106,6 +107,11 @@ class TypeMap(dict):
     def add_type(self, graphene_type):
         if inspect.isfunction(graphene_type):
             graphene_type = graphene_type()
+
+        # If type is a GraphQLType from graphql-core then return it immediately
+        if isinstance(graphene_type, GraphQLType):
+            return graphene_type
+
         if isinstance(graphene_type, List):
             return GraphQLList(self.add_type(graphene_type.of_type))
         if isinstance(graphene_type, NonNull):
--- a/graphene/types/schema.py
+++ b/graphene/types/schema.py
@@ -258,7 +258,8 @@ class TypeMap(dict):
             union_types = []
             for graphene_objecttype in graphene_type._meta.types:
                 object_type = create_graphql_type(graphene_objecttype)
-                assert object_type.graphene_type == graphene_objecttype
+                if hasattr(object_type, "graphene_type"):
+                    assert object_type.graphene_type == graphene_objecttype
                 union_types.append(object_type)
             return union_types
 
--- a/graphene/types/tests/test_type_map.py
+++ b/graphene/types/tests/test_type_map.py
@@ -1,3 +1,5 @@
+from textwrap import dedent
+
 from graphql.type import (
     GraphQLArgument,
     GraphQLEnumType,
@@ -270,3 +272,44 @@ def test_objecttype_with_possible_types():
     assert graphql_type.is_type_of
     assert graphql_type.is_type_of({}, None) is True
     assert graphql_type.is_type_of(MyObjectType(), None) is False
+
+
+def test_graphql_type():
+    """Type map should allow direct GraphQL types"""
+    MyGraphQLType = GraphQLObjectType(
+        name="MyGraphQLType",
+        fields={
+            "hello": GraphQLField(GraphQLString, resolve=lambda obj, info: "world")
+        },
+    )
+
+    class Query(ObjectType):
+        graphql_type = Field(MyGraphQLType)
+
+        def resolve_graphql_type(root, info):
+            return {}
+
+    schema = Schema(query=Query)
+    assert str(schema) == dedent(
+        """\
+        type Query {
+          graphqlType: MyGraphQLType
+        }
+
+        type MyGraphQLType {
+          hello: String
+        }
+    """
+    )
+
+    results = schema.execute(
+        """
+        query {
+            graphqlType {
+                hello
+            }
+        }
+    """
+    )
+    assert not results.errors
+    assert results.data == {"graphqlType": {"hello": "world"}}
--- a/graphene/types/tests/test_type_map.py
+++ b/graphene/types/tests/test_type_map.py
@@ -1,6 +1,7 @@
 from textwrap import dedent
 
 import pytest
+from graphql import parse, build_ast_schema
 from graphql.type import (
     GraphQLArgument,
     GraphQLEnumType,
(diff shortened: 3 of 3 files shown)


syrusakbary:
This PR has to be a bit better analyzed (both technically and non-technically).

On the tech side, it might be not as trivial to do. On the non-tech side is not as clear on what are the use cases (we talked about Strawberry integration, but I feel very strongly about adding type support inside Graphene).
Ideally Graphene should strive to solve the GraphQL use cases itself, I think this patch might lead to the opposite.

jkimbo:
@syrusakbary 

> On the tech side, it might be not as trivial to do.

Can you expand on that and give examples? I don't understand what you mean.

> On the non-tech side is not as clear on what are the use cases (we talked about Strawberry integration, but I feel very strongly about adding type support inside Graphene).
Ideally Graphene should strive to solve the GraphQL use cases itself, I think this patch might lead to the opposite.

The use case in my head is for incremental migration to either another library or to Graphene from another library. For example say you've written your server using just graphql-core but you want to start using Graphene because you prefer the API, you could replace the root Query and Schema with Graphene but still use the graphql-core types you already have without having to rewrite them all.

Regarding Strawberry: even if we added Python type support inside Graphene, Strawberry might have other features that people prefer and again this PR would allow users to incrementally adopt that library rather than having to completely rewrite it.

syrusakbary:
> Can you expand on that and give examples? I don't understand what you mean.

What happens when you provide a raw interface type to a graphene ObjectType. Should the fields be reflected on the graphene ObjectType? If now, how should we do it? Is not trivial.

> The use case in my head is for incremental migration to either another library or to Graphene from another library

That's precisely what I'm concerned of. If people are migrating from Graphene to other libraries it will mean that Graphene itself has not evolved as needed. This PR provides a patch that eventually allow such scenarios to happen.

jkimbo:
> What happens when you provide a raw interface type to a graphene ObjectType. Should the fields be reflected on the graphene ObjectType? If not, then how should we do it? That's what I mean saying that the tech side might be not as trivial to do.

Thanks that's very helpful, I was missing those use cases. The Graphene Interface type already checks that the interfaces are Graphene types which I think is a good thing and we should keep because of the technical challenge that you've mentioned. For Unions: it only needed a small change to support unions of a mixture of Graphene and GraphQL types. I've added tests for both cases to assert their behaviour.

> That's precisely what I'm concerned with. If people are migrating from Graphene to other libraries it will mean that Graphene itself has not evolved as their users expected. This PR provides a patch that eventually allow such scenarios to happen.
Or more in detail: while this PR is agnostic of 3rd party libraries, it provides a scape way (using other libs) rather than a proper solution (improving graphene) in case there are things to improve in the framework.

I strongly disagree with this reasoning. Yes in an ideal world Graphene would be all things to all people but that is not likely or desirable in my opinion. This change allows for interop with other libraries (both current and future) built on top of graphql-core that might provide radically different APIs or functionality than Graphene shouldn't attempt to incorporate. This change allows people to experiment without having to wait on Graphene to improve which makes the whole ecosystem better. IMO Graphene is an opinionated way of build GraphQL servers and we should allow developers to incorporate alternative opinions without forcing them to rewrite everything.

syrusakbary:
> Yes in an ideal world Graphene would be all things to all people but that is not likely or desirable in my opinion.

Why not likely? I agree that somethings like SDL-first schemas are not a priority for Graphene. However annotation-based schemas is an objective of Graphene.

I can't see any use case for a mixture of different styles mixed together (GraphQL-core plain types + Graphene + SDL based schemas, for example) other than transitioning from one to other, which I don't believe we should strive for.

AndrewIngram:
I've personally mixed SDL-first and code-first types when building GraphQL servers in other languages. I'm not big on SDL-first but do appreciate its ergonomics, but you often hit roadblocks when it comes to metaprogramming and you have to do awkward string-building functions.

But there's a larger philosophical point here, and I'm strongly in agreement with @jkimbo here. Every project exists because it set out to solve problems in a way that aligns with the author's (or authors') principles of how things should be done. Two projects could be interchangeable on a raw capability level, but still be vastly different in how they choose to tackle the problems. The idea that someone choosing to migrate away is a damning criticism of your project, is an expression of a flawed winner-takes-all worldview. Your project might do every someone needs, but they might just enjoy working with something else more. Rails and Django are largely interchangeable, but some people like Python more and some people like Ruby more.

As a suggestion to @syrusakbary, make it clear what your values and objectives with Graphene are (though i'd prefer if you avoided grand ones like being the only library in town), and then it'll be clearer for contributors at every level where they stand on each issue.

syrusakbary:
> The idea that someone choosing to migrate away is a damning criticism of your project, is an expression of a flawed winner-takes-all worldview. Your project might do everything someone needs, but they might just enjoy working with something else more. Rails and Django are largely interchangeable, but some people like Python more and some people like Ruby more.

I 100% agree with this. However, I don't see people not choosing Graphene as a criticism for the project. I see people not choosing Graphene because it's missing features that I think Graphene should have, as a failure of the project.

One of this clear scenarios is typing/annotations. I believe typing annotations are going to be critical for Python in the future, makes of an easier API, and as such I think Graphene should have this bundled into the library. I even wrote Graphene 2 with [typing in mind](https://github.com/graphql-python/graphene/issues/729#issuecomment-393673570).

As commented in this PR, If we want to add GraphQL interexchange support as a way for people to use other library that have typing support I see as a failure of Graphene as a library itself (since it's one of its library goals).

> As a suggestion to @syrusakbary, make it clear what your values and objectives with Graphene are (though i'd prefer if you avoided grand ones like being the only library in town), and then it'll be clearer for contributors at every level where they stand on each issue.

But I agree with your point, expectations should be clear. Perhaps they have not been clear enough, so let me be a bit clearer.
Here are the values of Graphene as a library:
* It should help developers to use GraphQL in Python in the easiest way possible (sorry if this is generic, but was one of my goals when creating Graphene)
* It should be true to the GraphQL spec
* It should be Pythonic
* It should be API-centric over maintainable
* It should be maintainable over performant
* It should try to not trade performance for readability
* It should be well tested
* It should be well documented
* It should aim to maximize of GraphQL usage in Python projects.

Constructing a schema from a SDL string is not pythonic. It deviates typing to other system and makes harder to glue things together.

Supporting python annotations is however quite pythonic, and I believe it solves two issues:
1. It helps having static typing in the project
2. It helps to create schemas easily

This PR was created with the initial intention to interexchange Graphene and Strawberry types. As having annotation support is one of the objectives of Graphene (since a few years ago, sadly it haven't been implemented), this PR collides with the first principle as: people will need to use two libraries, a non-cohesive syntax and more artifacts.

If someone has other real usage for interexchange GraphQL types that is not covered by what I commented, please let me know.
Hope this things makes things clear!

jkimbo:
@syrusakbary I've just pushed another commit to show how you could interchange GraphQL-core SDL parsing with Graphene types. It's another example of something that would allow people to incrementally migrate to Graphene from another paradigm without having to re-write everything and would not be possible without this change.

> This PR was created with the initial intention to interexchange Graphene and Strawberry types. As having annotation support is one of the objectives of Graphene (since a few years ago, sadly it haven't been implemented), this PR collides with the first principle as: people will need to use two libraries, a non-cohesive syntax and more artifacts.

That was the original use case that I had in mind but as I have demonstrated there are other cases that would benefit from this change. We can all agree that adding type support to Graphene is an objective of Graphene but blocking this change because you believe it will encourage people to migrate to a different library is just petty. I have no intention of encouraging people to interchange libraries, this change is just the bare minimum to make it possible so that people can experiment if they want.

syrusakbary:
> blocking this change because you believe it will encourage people to migrate to a different library is just petty

Oh no, is not that way. My main concern is that it might make us lazy as Graphene developers. If people have already a way to use typing within Graphene they will not look to other packages (for example), with this PR they will simply stop asking for that feature within Graphene which will led to us relaxing on adding the feature to the library (which I believe very strongly Graphene should have).

My main concern about the PR is that it discourages usage within the library adding a scape goat. I don't believe that's a good way of approaching future problems. The side concern I have is that I this is resulting more in a philosophical discussion than practical one. I haven't seen an issue in this repo where people are trying to intercommunicate different libraries.

I'm sorry but I feel very strong about this. If you want to discuss further I'm happy to jump on a Zoom/hangouts to clarify further :)

syrusakbary:
Closing the PR for now.

For future readers: if you have a use case of mixing Graphene types with other types for an existing project, please write here your use case so we can recalibrate the strategy.

Issue #1226, opened 2020-07-12
Proposal: Alternative API for mutations and subscriptions


jkimbo:
# The problem

Currently the mutation api is quite clunky and not very intuitive: https://docs.graphene-python.org/en/latest/types/mutations/

Often users get confused about the `mutate` function being a static method and that you need to define a separate `Mutation` class that extends `ObjectType` with each mutation in it. Also the `Output` field is not a very intuitive way of defining the return value of the `Mutation`.

See: 
https://github.com/graphql-python/graphene/issues/1163 
https://github.com/graphql-python/graphene/issues/810 
https://github.com/graphql-python/graphene-django/issues/759 
and others

Subscriptions in Graphene are currently a very new and experimental feature and the current API has some downsides as outlined here: https://github.com/graphql-python/graphene/pull/1107#issuecomment-657194024

# The solution

Implement a new decorator based API for both mutations and subscriptions and expose a new `mutations` and `subscriptions` argument to `Schema` that accepts a list of mutation and subscription functions. This has the benefit of not only being simpler but also making it super clear that the mutation/subscribe function is a static method.

### Mutations:
```python
from graphene import mutation, String, Schema

from .definitions import User, Query

@mutation(User, required=True, arguments={"name": String(required=True)})
def update_user(root, info, name):
    # Update the user...
    return User(name=name)

schema = Schema(query=Query, mutations=[update_user])
```

<details>
<summary>Equivalent functionality with the current API</summary>

```python
from graphene import Mutation, String, Schema, ObjectType, Field

class UpdateUser(Mutation):
    class Arguments:
        name = String(required=True)

    Output = User

    def mutate(root, info, name):
        # Update the user...
        return User(name=name)

class MyMutation(ObjectType):
    update_user = UpdateUser.Field(required=True)

schema = Schema(query=Query, mutation=MyMutation)
```
</details>

### Subscriptions:

```python
from graphene import subscription, Int, Schema

from .definitions import Query

@subscription(Int, required=True)
async def count_to_ten(root, info):
    count = 0
    while count < 10:
        count += 1
        yield count

schema = Schema(query=Query, subscriptions=[count_to_ten])
```

# Drawbacks

Using decorators like this is quite different to the current API so it would require a lot of documentation and education to get users to migrate to it (though the 2 APIs could coexist quite easily). Since there should only be 1 way of defining mutations we should try and make it as easy as possible for people to migrate.

# Alternatives

[...]

PR #1229, opened 2020-07-14
Rename subscribe method to EXPERIMENTAL_subscribe

jkimbo:
Because we haven't settled on a good API for subscriptions yet (see discussion here: https://github.com/graphql-python/graphene/pull/1107#issuecomment-657194024) this PR renames the `subscribe` method on `Schema` to `EXPERIMENTAL_subscribe` to make it clear that the API might change.

### DIFF
--- a/graphene/types/schema.py
+++ b/graphene/types/schema.py
@@ -475,7 +475,7 @@ class Schema:
         kwargs = normalize_execute_kwargs(kwargs)
         return await graphql(self.graphql_schema, *args, **kwargs)
 
-    async def subscribe(self, query, *args, **kwargs):
+    async def EXPERIMENTAL_subscribe(self, query, *args, **kwargs):
         document = parse(query)
         kwargs = normalize_execute_kwargs(kwargs)
         return await subscribe(self.graphql_schema, document, *args, **kwargs)
--- a/graphene/types/schema.py
+++ b/graphene/types/schema.py
@@ -476,6 +476,11 @@ class Schema:
         return await graphql(self.graphql_schema, *args, **kwargs)
 
     async def EXPERIMENTAL_subscribe(self, query, *args, **kwargs):
+        """Create a GraphQL subscription.
+
+        NOTE: this is an EXPERIMENTAL API and as such does not follow the normal semver release guarantees. Use it at
+        your own risk.
+        """
         document = parse(query)
         kwargs = normalize_execute_kwargs(kwargs)
         return await subscribe(self.graphql_schema, document, *args, **kwargs)
--- a/tests_asyncio/test_subscribe.py
+++ b/tests_asyncio/test_subscribe.py
@@ -26,7 +26,7 @@ schema = Schema(query=Query, subscription=Subscription)
 @mark.asyncio
 async def test_subscription():
     subscription = "subscription { countToTen }"
-    result = await schema.subscribe(subscription)
+    result = await schema.EXPERIMENTAL_subscribe(subscription)
     count = 0
     async for item in result:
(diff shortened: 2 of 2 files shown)


syrusakbary:
I think it would make sense to change `subscribe` to `EXPERIMENTAL_subscribe` if we foresee any future changes to the API, or are uncertain about the current exposed API.

However, using async iterators I think is the way to go for subscriptions, and I can't think of a better API/usage for it than the current one (note: the fact that we are now using `graphql-core>=3` it implies that async iterators should be supported).

Do you think there is uncertainty on the subscription API @jkimbo ? If so, what you think could change on the API?

syrusakbary:
Note: After checking #1107 a bit further, I believe we need to fix the API of the subscription to move from:

```python
    time_of_day = Field(String)

    async def subscribe_time_of_day1(root, info):
        try:
            count = 0
            while count < 10:
                yield {'time_of_day': datetime.now().isoformat()}
                await asyncio.sleep(1)
                count += 1
        except Exception as error:
            print(error)
```

To:

```python
    time_of_day1 = Field(String)

    async def subscribe_time_of_day1(root, info):
        try:
            count = 0
            while count < 10:
                yield datetime.now().isoformat()
                await asyncio.sleep(1)
                count += 1
        except Exception as error:
            print(error)
```

The change on the graphene side should be trivial :)

syrusakbary:
If we are able to do that, then I don't think we will really need an `EXPERIMENTAL_` prefix, unless someone proposes or can think of a better API.

jkimbo:
@syrusakbary your example here: https://github.com/graphql-python/graphene/pull/1229#issuecomment-659939845 is what I was referring to regarding changing the API. Also as I outlined in https://github.com/graphql-python/graphene/pull/1107#issuecomment-657194024 calling subscribe with multiple fields will only execute the first resolver so we should add validation to the call. Finally the API is not documented so I don't think it's ready for public use.

I would like to get v3 released as soon as possible and it's already a very big release so I think it makes sense to not include subscriptions as part of it but finish the API changes and release it as part of v3.1. I would also like to come to a conclusion on #1226

syrusakbary:
I'm happy to merge this PR, BUT... I actually think the current subscription mechanism have to be improved before releasing Graphene 3.0.

With this I mean that I believe we should have this API working (note that is not returning dictionaries):

```python
    time_of_day = Field(String)

    async def subscribe_time_of_day(root, info):
        try:
            count = 0
            while count < 10:
                yield datetime.now().isoformat()
                await asyncio.sleep(1)
                count += 1
        except Exception as error:
            print(error)
```

jkimbo:
The whole point of this change is so that the v3 release can be unblocked without having to address the api changes right now (by making it clear that subscription support is still experimental). V3 already has a lots of new stuff that would be beneficial to release and has been in development for almost a year now. I don't want to hold it up any more. 

If you want to fix the api issues first then go ahead @syrusakbary but I don't have time to do that.

syrusakbary:
This PR shouldn't be any longer needed once we merge #1235.

Note: I do like to use prefixes (such as "experimental" or "dangerously" -like in React to set the element HTML directly). But I do like to use them in specific cases:
* Using `experimental` when the API itself is not clear (meaning: when you need more learnings as a package maintainer to assure it's stability)
* Using `dangerous` (or `unsafe` like in Rust) when a method is high-likely to be dangerous for the user.

However, this PR is adding the `EXPERIMENTAL` prefix mainly because we didn't spent the time to polish the API for subscriptions, not because we didn't know how to do it or what the final API would look like. So once #1235 is merged we should close this :)

Issue #1233, opened 2020-07-21
Why SubclassWithMeta.__init_subclass__ terminates the super() chain


ruohola:
Is there a great reason why `graphene.utils.subclass_with_meta.SubclassWithMeta.__init_subclass__`, doesn't call `super().__init_subclass__()`? 
https://github.com/graphql-python/graphene/blob/master/graphene/utils/subclass_with_meta.py#L21-#L46

The issue with this IMO is the following:

Class `Foo` has a custom `__init_subclass`. Class `Bar` derives from `Foo` and e.g. `graphene.types.mutation.Mutation`.
Now if `Foo` is accidentally placed before `Mutation` in the inheritance list, `Foo`'s `__init_subclass` will never be run.

PR #1234, opened 2020-07-21
Don't terminate the `super()` chain in `SubclassWithMeta.__init_subclass__`

ruohola:
Resolves https://github.com/graphql-python/graphene/issues/1233

### DIFF
--- a/graphene/utils/tests/test_subclass_with_meta.py
+++ b/graphene/utils/tests/test_subclass_with_meta.py
@@ -1,0 +1,63 @@
+from ..subclass_with_meta import SubclassWithMeta
+
+
+def test_init_subclass_calls_super():
+    class DefinesInitSubclass:
+        subclass_init_count = 0
+
+        def __init_subclass__(cls, **kwargs):
+            DefinesInitSubclass.subclass_init_count += 1
+            super().__init_subclass__(**kwargs)
+
+    class InheritsAsFirst(SubclassWithMeta, DefinesInitSubclass):
+        pass
+
+    class InheritsAsLast(DefinesInitSubclass, SubclassWithMeta):
+        pass
+
+    assert DefinesInitSubclass.subclass_init_count == 2
+
+
+def test_kwargs_are_consumed():
+    class DefinesInitSubclass:
+        passed_kwargs = []
+
+        def __init_subclass__(cls, **kwargs):
+            if kwargs:
+                DefinesInitSubclass.passed_kwargs.append(kwargs)
+            super().__init_subclass__(**kwargs)
+
+    class InheritsAsFirst(SubclassWithMeta, DefinesInitSubclass, is_consumed="foo"):
+        pass
+
+    class InheritsAsLast(DefinesInitSubclass, SubclassWithMeta, not_consumed="bar"):
+        pass
+
+    assert DefinesInitSubclass.passed_kwargs == [{"not_consumed": "bar"}]
+
+
+def test_meta_is_deleted_and_props_passed():
+    class DefinesInitSubclassWithMeta(SubclassWithMeta):
+        passed_meta_options = []
+
+        @classmethod
+        def __init_subclass_with_meta__(cls, **options):
+            if options:
+                DefinesInitSubclassWithMeta.passed_meta_options.append(options)
+            super().__init_subclass_with_meta__()
+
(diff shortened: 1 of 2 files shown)


jkimbo:
@syrusakbary any insight into why super is not being called here?

syrusakbary:
Can you add a test @ruohola? It will make a bit easier what you are trying to fix, and it's side effects.

Thanks!

ruohola:
@syrusakbary Added some texts, pretty contrived ones, but I think they should be useful.

graphene/utils/tests/test_subclass_with_meta.py
```diff
@@ -0,0 +1,63 @@
...
+            DefinesInitSubclass.subclass_init_count += 1
+            super().__init_subclass__(**kwargs)
+
+    class InheritsAsFirst(SubclassWithMeta, DefinesInitSubclass):
+        pass
+
+    class InheritsAsLast(DefinesInitSubclass, SubclassWithMeta):
+        pass
+
+    assert DefinesInitSubclass.subclass_init_count == 2
```
ruohola, line 18: This would fail with`assert 1 == 2` without the added `super()` call here:

https://github.com/graphql-python/graphene/blob/9d96fc4d86d4114cc9bdac301d4c7bc58ac4f5a7/graphene/utils/subclass_with_meta.py#L52

graphene/utils/tests/test_subclass_with_meta.py
```diff
@@ -0,0 +1,63 @@
...
+def test_kwargs_are_consumed():
+    class DefinesInitSubclass:
+        passed_kwargs = []
+
+        def __init_subclass__(cls, **kwargs):
+            if kwargs:
+                DefinesInitSubclass.passed_kwargs.append(kwargs)
+            super().__init_subclass__(**kwargs)
+
+    class InheritsAsFirst(SubclassWithMeta, DefinesInitSubclass, is_consumed="foo"):
```
syrusakbary, line 30: Is not clear to me what happens with the `is_consumed="foo"` meta variable.

I'm a bit concerned about the inconsistency between `InheritsAsFirst` and `InheritsAsLast`

ruohola: Well, that inconsistency is there even without this PR. I agree that it's a bit messy, maybe we could do:

```
try:
    super().__init_subclass__(**meta_options)
except TypeError:
    super().__init_subclass__()
```

Without the try-ecxept this will fail when `super()` is `object`, since it's `__init_subclass__` doesn't take any arguments.

ruohola: > ```
> try:
>     super().__init_subclass__(**meta_options)
> except TypeError:
>     super().__init_subclass__()
> ```

Okey, this is actually a bad way to do it, and can hide accidentally suppress unwanted errors. I think that it's fine that the `__init_subclass__` consumes all of the kwargs, since it really cannot know if the next class in the MRO is able to receive the kwargs.

syrusakbary: I need more time to think about this and come to a conclusion. Thanks for the analysis @ruohola

ruohola: Yeah, I truly think this is how the method should behave. The docs tell us https://docs.python.org/3/reference/datamodel.html#object.__init_subclass__, that we are supposed to take the away needed kwargs and then continue the super chain. Since `SubclassWithMeta` is specifically made so that it takes use of all of the passed kwargs, we should take all of them away and just call the bare `super().__init_subclass__`.

ruohola:
So what do you think, could this be merged? 🧐

ruohola:
Sorry, my git scripts messed up ^^

ruohola:
@syrusakbary So, have you made a decision?

graphene/utils/tests/test_subclass_with_meta.py
ruohola: @syrusakbary So, have you made a decision?

ruohola: @syrusakbary So, what do you think?

ruohola: @syrusakbary @jkimbo So, what do you think?

ruohola:
w/e 🤷‍♂️

PR #1235, opened 2020-07-27
Subscription revamp

syrusakbary:
The implemented subscription mechanism was having some issues, namely:
1. It was not validating queries / throwing errors on "unparseable" queries
2. It's API for yielding data for a subscription was not super good (you have to do `yield {"the_field_name": data}` rather than `yield data`.

This PR improves the subscription mechanism to a "production-grade" quality (par to graphql-core and graphq-js counterparts), so we can encourage a proper use for Graphene users while avoiding paths such as discouraging use via renaming of the subscribe method (see #1229).

This PR also merges the `tests_asyncio` directory into the normal tests (we have this on previous versions of Graphene to make sure we support both non-asyncio and asyncio environments, which is no longer needed since we can presume async support is always present thanks to Python 3).


graphene/types/field.py
```diff
@@ -114,5 +115,24 @@ def __init__(
...
-    def get_resolver(self, parent_resolver):
+    get_resolver = None
+
+    def wrap_resolve(self, parent_resolver):
+        """
+        Wraps a function resolver, using the ObjectType resolve_{FIELD_NAME}
+        (parent_resolver) if the Field definition has no resolver.
+        """
+        if self.get_resolver is not None:
+            warn_deprecation(
```
jkimbo, line 126: Why the name change?

syrusakbary: It's much clearer what it does. It wraps the function provided (or defined in self).

Issue #902, opened 2019-02-06
Error handling


sebastiandev:
We have decided to approach error handling like suggested in many blogs, by having errors defined as fields rather than raising exceptions.

The thing with this is that we need to make sure we catch every exception that we want to handle and somehow keep track of them to be able to inform them.

So we came up with:
```python

class BaseObjectType(graphene.ObjectType):

    class Meta:
        abstract = True

    @classmethod
    def __init_subclass_with_meta__(
        cls,
        interfaces=(),
        possible_types=(),
        default_resolver=None,
        _meta=None,
        **options
    ):
        super().__init_subclass_with_meta__(
            interfaces, possible_types, default_resolver, _meta, **options)

        for f in cls._meta.fields:
            if f in ['errors']:
                continue

            resolver_name = "resolve_{}".format(f)
            if hasattr(cls, resolver_name):
                setattr(cls, resolver_name, catch_errors(getattr(cls, resolver_name)))


class BaseResponse(BaseObjectType):

    class Meta:
        abstract = True

    errors = graphene.String()

    @staticmethod
    def resolve_errors(root, info, **kwargs):
        operation_name = info.path[0]
        error_key = f"{operation_name}_errors"
   
        if not root.errors and error_key in info.context.errors:
            root.errors = ",".join(info.context.errors[error_key])

        return root.errors
```

`catch_errors` just populates `info.context.errors` for each operation 

While implementing this approach for error handling I have also found an issue when resolving the error field. I have a base class that extends ObjectType by iterating over the resolvers and wrapping them to n catch any errors and setting an errors dict in the info.context.

Problem is that when resolving a type field that throws an exception, I do catch it and set the error, but the error field was already processed/resolved. Is there a way to force a particular field to be processed at the end or to force it to update? I need to make sure every resolver runs before the error field is resolver so I make sure all errors are caught.

BossGrand:
This is possible! You need to wrap the resolver like such

```

def wrap_resolver(cls, resolver):

    def wrapped_resolver(root, info, **args):
        try:
            with transaction.atomic():
                return resolver(root, info, **args)

        except Exception as e:
            return cls(errors=str(e))

    return wrapped_resolver


class CustomMutation(Mutation):

    class Meta:
        abstract = True


    @classmethod
    def __init_subclass_with_meta__(cls, resolver=None, **options):

        if not resolver:
            mutate = getattr(cls, 'mutate', None)
            assert mutate, 'All mutations must define a mutate method in it'
[...]

PR #1255, opened 2020-08-18
Add UnforgivingExecutionContext

AlecRosenbaum:
Resolves: https://github.com/graphql-python/graphene/issues/902

I'm not sure what the best place to add this execution context is. I'm happy to change naming and where this is defined to whatever is most appropriate.

Assuming this approach looks good, this probably also warrants some updates to the docs before merging.

Note that I have not added this as the default execution context since that would be a pretty breaking change. If a breaking change is ok, I'll update this PR.

In the project where I've taken a similar approach, I also added an `assert graphl.version == '3.1.1'` as a reminder to check the underlying implementation for changes when bumping versions. It might be worth adding a similar check here as well

### DIFF
--- a/graphene/types/schema.py
+++ b/graphene/types/schema.py
@@ -319,7 +319,7 @@ class TypeMap(dict):
                     )
                 subscribe = field.wrap_subscribe(
                     self.get_function_for_type(
-                        graphene_type, f"subscribe_{name}", name, field.default_value,
+                        graphene_type, f"subscribe_{name}", name, field.default_value
                     )
                 )
 
@@ -405,7 +405,7 @@ class UnforgivingExecutionContext(ExecutionContext):
     """
 
     def resolve_field_value_or_error(
-        self, field_def, field_nodes, resolve_fn, source, info,
+        self, field_def, field_nodes, resolve_fn, source, info
     ):
         """Resolve field to a value or an error.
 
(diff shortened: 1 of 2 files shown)


AlecRosenbaum:
@jkimbo I don't think I have access to request a review. What's the process from here?

AlecRosenbaum:
@jkimbo @syrusakbary I've updated this PR with fixes for an issue I found when using this in a different project.

What are the next steps for either accepting or rejecting this change?

zbyte64 (approved):
I like the interface and the results.
:rocket:

jkimbo (approved):
Sorry for the delay @AlecRosenbaum . This looks great thanks!

jkimbo:
Released as [v3.0.0b6](https://github.com/graphql-python/graphene/releases/tag/v3.0.0b6)

coler-j:
@AlecRosenbaum any docs for how to use this?

zbyte64:
> @AlecRosenbaum any docs for how to use this?

No docs yet but I pulled this from the unit tests:

```
           schema.execute(
                f"query {{ {field} }}",
                execution_context_class=UnforgivingExecutionContext,
            )
```

AlecRosenbaum:
Sorry @coler-j, I meant to respond to this but got distracted.

The unit test example above is probably the best example of how to use it. In the project I use this, we do some other custom stuff that wouldn't show the usage very well.

weilu:
This no longer works with graphene 3.1.5 as `resolve_field_value_or_error` is removed. The fix is rather simple: https://github.com/weilu/graphene/commit/bec21917d5c5e63e6f6d500843270f92fbaaa771, but I haven't had time to look into tests so I'm just leaving this here to save time for those who have the same issue.

PR #1282, opened 2020-11-14
Fixing a Type Warning as per #1100

### LINKED ISSUE #1100: PyCharm warning: Expected type 'ObjectType', got 'Type[Query]' instead
```py

    # ...

schema = Schema(query=Query)

```

image

Related to #814 which was automatically closed due to staleness.

Any ideas?

evank28:
Fixing a Type Warning that resolves #1100 for PyCharm IDE users.

### DIFF
--- a/graphene/types/schema.py
+++ b/graphene/types/schema.py
@@ -1,5 +1,6 @@
 import inspect
 from functools import partial
+from typing import Type
 
 from graphql import (
     default_type_resolver,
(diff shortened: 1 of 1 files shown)


skewty:
should it be `Optional[Type[ObjectType]]`?  `None` seems to be an acceptable value.

erikwrede:
We should fix all type hints and switch to this: 
https://github.com/tox-dev/sphinx-autodoc-typehints

@skewty is right, `Query` should be optional. See `types/schema.py:107` and `graphql-core/type/schema.py:95`. Closing this, as single-line type hint corrections don't solve the overall problem. PRs should be at least on a per-file basis to ensure we are not skimming through every file twice.

PR #1284, opened 2020-11-17
Support leading underscore in to_camel_case

JMSwag:


JMSwag:
https://github.com/graphql-python/graphene-mongo/issues/153

PR #1297, opened 2021-01-05
fix: test client is updated to be asyncio and not promises

### LINKED ISSUE #1296: Test Client is still promise based
* **What is the current behavior?**
When trying to use the test client in unit tests, it is still based on Promises which is no longer imported.

* **What is the expected behavior?**
Test client needs to be updated to be asyncio.

* **What is the motivation / use case for changing the behavior?**
Promises are no longer part of the library.

wakemaster39:
resolves #1296


erikwrede (changes requested):
Hey, sorry for the late reply. Could you please add a test case for execute_async to maintain code coverage?

erikwrede:
Superseded by #1476 which has been merged. Thanks for getting this started!

PR #1300, opened 2021-01-06
Argument default value should be undefined

### LINKED ISSUE #1195: Default value for Argument's should be Undefined
When upgrading from 2 to 3, the default value of Argument's has changed from undefined to none which changes the schema. An example from `test_relay_connection`: 

```
type Query {
  letters(before: String = null, after: String = null, first: Int = null, last: Int = null): LetterConnection
  connectionLetters(before: String = null, after: String = null, first: Int = null, last: Int = null): LetterConnection
  asyncLetters(before: String = null, after: String = null, first: Int = null, last: Int = null): LetterConnection
  node(
    """The ID of the object"""
    id: ID!
  ): Node
}
```

which should be:

```
type Query {
  letters(before: String, after: String, first: Int, last: Int): LetterConnection
  connectionLetters(before: String, after: String, first: Int, last: Int): LetterConnection
  asyncLetters(before: String, after: String, first: Int, last: Int): LetterConnection
  node(
    """The ID of the object"""
    id: ID!
  ): Node
}
```

Happy to submit a PR to fix if this is not intended behavior.

wakemaster39:
resolves #1195

This stops `null` from being populated in the introspection query as mentioned in #1195. This also solves the problem of `None` being passed into the kwargs of a `mutate` call if no value was passed in by the end user.

### DIFF
--- a/graphene/types/argument.py
+++ b/graphene/types/argument.py
@@ -1,5 +1,7 @@
 from itertools import chain
 
+from graphql import Undefined
+
 from .dynamic import Dynamic
 from .mountedtype import MountedType
 from .structures import NonNull
@@ -41,7 +43,7 @@ class Argument(MountedType):
     def __init__(
         self,
         type_,
-        default_value=None,
+        default_value=Undefined,
         description=None,
         name=None,
         required=False,
@@ -102,9 +104,7 @@ def to_arguments(args, extra_args=None):
             raise ValueError(f'Unknown argument "{default_name}".')
 
         arg_name = default_name or arg.name
-        assert (
-            arg_name not in arguments
-        ), f'More than one Argument have same name "{arg_name}".'
+        assert arg_name not in arguments, f'More than one Argument have same name "{arg_name}".'
         arguments[arg_name] = arg
 
     return arguments
--- a/graphene/types/argument.py
+++ b/graphene/types/argument.py
@@ -104,7 +104,9 @@ def to_arguments(args, extra_args=None):
             raise ValueError(f'Unknown argument "{default_name}".')
 
         arg_name = default_name or arg.name
-        assert arg_name not in arguments, f'More than one Argument have same name "{arg_name}".'
+        assert (
+            arg_name not in arguments
+        ), f'More than one Argument have same name "{arg_name}".'
         arguments[arg_name] = arg
 
     return arguments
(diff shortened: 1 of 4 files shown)


wakemaster39:
I am unsure which of the two options you want me to do to fix the unit tests.

1, set default values for the failing tests to be None.
2, remove the no longer passed in values from the assertion statements.

DonQueso89:
> I am unsure which of the two options you want me to do to fix the unit tests.
> 
> 1, set default values for the failing tests to be None.
> 2, remove the no longer passed in values from the assertion statements.

thumbs up for this PR! 

I would say remove the no longer passed in values from the assertion right? We're testing whether the value is absent if its not passed in

graphene/types/argument.py
```diff
@@ -102,9 +104,7 @@ def to_arguments(args, extra_args=None):
             raise ValueError(f'Unknown argument "{default_name}".')
 
         arg_name = default_name or arg.name
-        assert (
-            arg_name not in arguments
-        ), f'More than one Argument have same name "{arg_name}".'
+        assert arg_name not in arguments, f'More than one Argument have same name "{arg_name}".'
```
zbyte64, line 107: I think this is causing the test to fail. It applies black formatting to the code an fails if the file changes. This is where it cites the changed file: https://travis-ci.org/github/graphql-python/graphene/jobs/753187127#L243

zbyte64 (changes requested):
See the commit to fix the test.

wakemaster39:
Ok I finally got around to cleaning up the changes that moving to Undefined makes to the unit tests. This should be good to go now.

zbyte64 (approved):
:+1:

daynejones (approved):

lovetoburnswhen:
+1 this seems like a critical feature

SteinRobert:
Does this PR need any more work? I'd be happy to help, otherwise I would really appreciate this to make it into the next release 😃

pavelion (approved):

frankchen211:
any update?

artemijan:
I believe the best way is just fork this repo, right? I need to differentiate null values with not provided in my app. This thread is about 1 year

SteinRobert:
@Cito @jkimbo you guys probably have a lot of work on your hands, thank you so much for caring about the project!
Just wanted to check in whether this change is something you consider accepting or if it needs any more work. I'd surely help making any more changes if needed.

erikwrede:
Hey everyone, will check this out in detail later. 
This might be a breaking change (breaking the changes from v2-> v3), but I believe releasing a new minor version for this will be sufficient.

PR #1304, opened 2021-01-15
Add support for interfaces on interfaces

AlecRosenbaum:
`graphql-core` [already supports specifying interfaces on interfaces](https://github.com/graphql-python/graphql-core/blob/7d826f0ec0a447cb869dfb891a755dbe8ea9b66f/src/graphql/type/definition.py#L819). This change adds that support to graphene.

For example, the interfaces from one of the added test cases would look like this in a printed schema:
```graphql
interface FooInterface {
  foo: String
}

interface BarInterface implements FooInterface {
  foo: String
  bar: String
}
```

I've tested these changes with a private schema and it seems to work as expected. There's some trickiness around getting graphql-core to resolve types properly but it's definitely doable.

### DIFF
--- a/graphene/types/interface.py
+++ b/graphene/types/interface.py
@@ -5,11 +5,12 @@ from .utils import yank_fields_from_attrs
 # For static type checking with Mypy
 MYPY = False
 if MYPY:
-    from typing import Dict  # NOQA
+    from typing import Dict, Iterable, Type  # NOQA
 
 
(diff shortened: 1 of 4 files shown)


zbyte64 (approved):
Definitely a nice to have feature :rocker:

erikwrede (approved):
Sorry for the long wait, this looks great!

PR #1308, opened 2021-02-03
Refactor project

kumaraditya303:
# Aim for PR :rocket:

- [x] Remove redundant build files like `isort.cfg` etc
- [x] Automate versioning with setuptools_scm hence no need to define `__version__` manually
- [x] Support for python 3.9 , latest stable python
- [x] Test on python nightly version 
- [x] Remove `setup.py test` command as it is deprecated https://tox.readthedocs.io/en/latest/example/basic.html#integration-with-setup-py-test-command
- [x]  Remove `dataclasses`  backport and auto install it on python version <3.7
More to be added as done.


ulgens:
What does this PR aim for? Updating build files and deleting some necessary stuff (i don't know why you thought they should be deleted) be two different PRs in my opinion.

ulgens:
@kumaraditya303 Each item on that list can be a different PR and even when that's the case, contents of them should be heavily discussed. I would personally recommend not to spend your time on a big "covers everything" PR like this. It's almost certain this PR will never get merged.

kumaraditya303:
Ok, so I should split the pr into multiple PRs. The main aim for this pr
was to remove all the unnecessary code and bring use new packages such as
setuptools_scm etc.

Issue #1303, opened 2021-01-15
Root Value example doesn't work


zegelin:
The [Root Value](https://docs.graphene-python.org/en/latest/execution/execute/#root-value) example in the documentation:

```
from graphene import ObjectType, Field, Schema

class Query(ObjectType):
    me = Field(User)

    def resolve_user(root, info):
        return {'id': root.id, 'firstName': root.name}

schema = Schema(Query)
user_root = User(id=12, name='bob')
result = schema.execute(
    '''
    query getUser {
        user {
            id
            firstName
            lastName
        }
    }
    ''',
    root=user_root
)
assert result.data['user']['id'] == user_root.id
```

doesn't work. Throws `NameError: name 'User' is not defined`.

What's not clear is:

- Where `User` is defined -- is it defined somewhere else in the documentation? (hyperlink/inline)
- The `Query` field is named `me` yet the resolver function is named `resolve_user`.
- The executed query asks for `user` (not `me`).
- The created `User` object (assigned to `user_root`) is constructed to have a `name` field, yet the query asks for the `firstName` and `lastName` fields. Sure, `resolve_user` returns a dict with `id`, and `firstName`, but this adds to the confusion.

Having various combinations of `name`, `firstName` and `lastName` is confusing, especially in an example that doesn't work.

PR #1310, opened 2021-02-07
Root Value example doesn't work #1303

sarathak:
example code updated


ekampf:
@sarathak this seems like a duplicate of https://github.com/graphql-python/graphene/pull/1327 ?

PR #1311, opened 2021-02-22
Raise error when arguments conains only type annotations

EmrysMyrddin:
Hello,

I'm a new user of graphene and I found it pretty cool for now !

But I came across a little annoying problem: I keep using `:` instead of `=` when defining types...

I suggest to add an error to detect this and give an error explicitly saying to the developer what is wrong with his use of the library.

This is a first approach, but we should probably make this a more general check, since it can occur in any graphql type definition, not only Arguments.

If you have suggestion about how to apply this more generaly, or if you think it's ok to make ad-hok checks, please comment :-)

### DIFF
--- a/graphene/types/argument.py
+++ b/graphene/types/argument.py
@@ -75,6 +75,10 @@ def to_arguments(args, extra_args=None):
     from .field import Field
     from .inputfield import InputField
 
+    if type(args) == dict and len(args) == 1 and '__annotations__' in args:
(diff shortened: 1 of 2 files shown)


erikwrede:
Closing this as strawberry solves this problem.

Currently there is not enough time to add such a feature to graphene and ensure that it stays well-maintained. If you believe that this is important and want to take care of the feature long-term, please open a new discussion.

PR #1048, opened 2019-08-09
Graphene v3 following v3 graphql-core

mvanlonden:
Graphene compatibility with graphql-core v3

Order of operations:
1. ~Approve and merge 988 into next/master https://github.com/graphql-python/graphene/pull/988~
2. ~Create a v2 branch from master~
3. ~PR and merge next/master into master https://github.com/graphql-python/graphene/pull/1048~
4. Cut a v3.0.0a0 release from master

PR #1326, opened 2021-04-17
Use argument's `default_value` regardless if the input field is required

minhtule:
The [GraphQL spec](https://github.com/graphql/graphql-spec/pull/418) allows a required argument to have a default value. In fact, [Graphene v2 supports this behavior](https://github.com/graphql-python/graphene/blob/v2/graphene/types/typemap.py#L294). However, for some reason, this is not allows in [Graphene v3](https://github.com/graphql-python/graphene/pull/1048/files#diff-f066ce1b84329fbfb7cd0e5d5027a51f68e14677754aab172fbc91dea42c519dR360-R362).

### DIFF
--- a/graphene/types/schema.py
+++ b/graphene/types/schema.py
@@ -26,7 +26,6 @@ from graphql import (
     GraphQLObjectType,
     GraphQLSchema,
     GraphQLString,
-    Undefined,
 )
 from graphql.execution import ExecutionContext
 from graphql.execution.values import get_argument_values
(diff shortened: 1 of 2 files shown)


ekampf:
Just adding this description from spec:
```
NEW: Non-null arguments and input object fields are no longer required to be provided a value/variable if that argument also supplies a default value.
```

ekampf (approved):

PR #1328, opened 2021-04-21
Include extra data in Meta in _meta

adrianschneider94:
I'm building an authorization middleware and I want to include functions/data in the Meta class to store the relevant information in the class itself.
However, all custom args in Meta are dropped at the moment.
I think there shouldn't be an issue to include other args in an extra attribute of _meta.

### DIFF
--- a/graphene/types/base.py
+++ b/graphene/types/base.py
@@ -43,6 +43,7 @@ class BaseType(SubclassWithMeta):
             return
         _meta.name = name or cls.__name__
         _meta.description = description or trim_docstring(cls.__doc__)
+        _meta.extra = _kwargs
         _meta.freeze()
         cls._meta = _meta
         super(BaseType, cls).__init_subclass_with_meta__()


adrianschneider94:
Hey, could anyone have a look on this?
It's a very tiny change and does not break anything...

adrianschneider94:
Now the tests are broken... I have no idea, why...

graphene/types/base.py
```diff
@@ -43,6 +43,7 @@ def __init_subclass_with_meta__(
             return
         _meta.name = name or cls.__name__
         _meta.description = description or trim_docstring(cls.__doc__)
+        _meta.extra = _kwargs
```
erikwrede, line 46: I think the `extra` field should be an explicit class attribute of the `BaseOptions` class so that it can be properly documented. Otherwise users will be confused about the meaning or the purpose/won't find it at all.

erikwrede (changes requested):

PR #1337, opened 2021-05-20
Using graphene < 3 with python > 3.8

cglacet:
This aims at allowing to use graphene < 3 with python > 3.8.


cglacet:
Can't seem to make python < 3 work, but I guess it's not related because it doesn't work on previous version either. Do you really want to keep python 2 support?

memark:
What's the status on this fix?

erikwrede:
Should be fixed by #1350

PR #1342, opened 2021-06-20
Create Either Argument decorator

Ian-Gabaraev:
Adds a decorator for handling situations when a query expects different arguments, but can only work with one. For instance:

A query named `userByAttr` can retrieve a user by their email or username, but not both. In this case, we define the query as follows:

```
from graphene.utils.decorators import either

@either(["username", "email"])
def resolve_user_by_attr(root, info, username=None, email=None):
        ...
```

If both arguments were passed, the query will raise `TooManyArgs`.
If no expected argument was passed, the query will raise `TooFewArgs`

### DIFF
--- a/graphene/utils/decorators.py
+++ b/graphene/utils/decorators.py
@@ -15,7 +15,7 @@ def either(expected):
             if not any(arg_list):
                 raise TooFewArgs("Too few arguments, must be either of: " + ",".join(expected))
 
-            if all(arg_list) or arg_list.count(True) > 1:
+            if arg_list.count(True) > 1:
                 raise TooManyArgs("Too many arguments, must be either of: " + ",".join(expected))
 
             return func(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


erikwrede (changes requested):
Thanks for the PR; sorry for the long wait!
While this is a useful QoL improvement for some users, I am worried that this change would promote unhealthy design choices. The strength of GraphQL is its typed schema. Creating a mutation that implicitly requires different combinations of fields may lead to user confusion.

We should definitely pick this up since the `@oneOf` feature will probably be added to spec, as that would resolve the issues of implicitness: https://github.com/graphql/graphql-spec/pull/825

However, this PR wouldn't support all use cases of `@oneOf`, e.G. nested `@oneOf`s inside of input types. I think the logic for this would better fit into the `Argument` handling.

If someone can point out additional examples of why this PR is still helpful, I am happy to reconsider. In the meantime, I think the technical implementation of this feature needs some extra discussion.

PR #1349, opened 2021-07-14
Add github actions to run linting and tests

jkimbo:


setup.py
```diff
@@ -49,6 +49,7 @@ def run_tests(self):
     "pytest-benchmark",
     "pytest-cov",
     "pytest-mock",
+    "fastdiff==0.2.0",
```
wodCZ, line 52: Why is this pinned? Could use a comment for future reference.

wodCZ:
Comparing with the travis.yml, I think `coveralls` step is missing? https://github.com/graphql-python/graphene/blob/master/.travis.yml#L12-L14

Not experienced with the tooling, but looks good to me other than that 👍

setup.py
jkimbo: Comment added.

jkimbo:
@wodCZ I don't think coveralls is that important so I'm going to skip it in favour of getting this merged sooner.

PR #1357, opened 2021-08-13
add support for query validation

### LINKED ISSUE #907: Protection against malicious queries
Hey. I was looking for a lot of information on how to protect against malicious requests, and as a result I found several common approaches:

- query cost or resource limitations (https://developer.github.com/v4/guides/resource-limitations/)
- limiting query depth (https://sangria-graphql.org/learn/#limiting-query-depth)
- query whitelisting
- hide introspection for production mode (this is not common solution, but sounds good)

All this approaches we can implement by meddleware and custom backend. But it will be cool if this solution is out of the box.

Also, graphene don't have information about security into the documentation.

ps: I can try to help, but if u have no reason why this is a bad idea.

codebyaryan:
adds a few validation rules which would be very useful in production environments. Documentation regarding validators have been added, and new test cases have also been made (which pass).

Similar validation rules are provided out of the box in other libraries like strawberry graphql and ariadne, so it would be nice for graphene users to get a similar experience.

validation rules which have been added:
- [x] depth limit validator
- [x] disable introspection

allows passing the validation rules to the `validate` method from graphql-core. Huge thanks to @jkimbo for his work on the depth limit validator, which was a port from https://github.com/stems/graphql-depth-limit

also fixes #907 
also aims to fix CI, replacing travis CI with github actions


codebyaryan:
@syrusakbary could you please review this if you have the time? It would be really helpful to use in production environments. Please let me know if any changes should be made, I'll do it! Thanks a lot!

syrusakbary:
I'm liking this PR! Could it be possible if we fix CI as part of the PR? (replacing Travis with Github CI should be good!)

codebyaryan:
I will try to fix it ASAP! Thanks a lot for the reply!

codebyaryan:
@syrusakbary please check the new workflow I pushed in (also deleted the .travis.yml file). If any changes need to be done, please let me know! The `PYPI_API_TOKEN` secret needs to be set to automatically publish to PyPi, I think.

graphene/validation/disable_introspection.py
```diff
@@ -0,0 +1,19 @@
...
+        field_name = node.name.value
+        if not is_introspection_key(field_name):
+            return
+
+        self.report_error(
+            GraphQLError(
+                f"Cannot query '{field_name}': introspection is disabled.",
+                node,
+            )
+        )
```
syrusakbary, line 19: I think is better the following logic, as it's a bit simpler and easier to follow:
```suggestion
        if is_introspection_key(field_name):
            self.report_error(
                GraphQLError(
                    f"Cannot query '{field_name}': introspection is disabled.",
                    node,
                )
            )
```

syrusakbary:
The PR looks great, although tests are not passing (it seems to be on `master`).
If you could help solving those last issues then we should be good to merge!

codebyaryan:
Sure, I've managed to find out the problem. There have been a few changes in graphql core lately. The ExecutionError provided by graphql-core doesn't eat up exceptions now.

You can check here for reference.
https://github.com/graphql-python/graphql-core/blob/afd03a058d04d12484dd920956d4611c50bf2f86/src/graphql/execution/execute.py#L580-L656

Therefore, we could fix the failing tests by removing UnforgivingExecutionContext and related tests (we dont need it anymore)

codebyaryan:
@AlecRosenbaum what do you think about this?

AlecRosenbaum:
> The ExecutionContext class provided by graphql-core doesn't eat up exceptions now.

If this was true, then the test suite should pass without specifying a custom context, right?

codebyaryan:
> > The ExecutionContext class provided by graphql-core doesn't eat up exceptions now.
> 
> If this was true, then the test suite should pass without specifying a custom context, right?

Yep, could you please check my commits?

graphene/types/tests/test_schema.py
```diff
@@ -69,115 +69,3 @@ def test_schema_requires_query_type():
     assert len(result.errors) == 1
     error = result.errors[0]
     assert error.message == "Query root type must be provided."
-
-
-class TestUnforgivingExecutionContext:
```
AlecRosenbaum, line 74: The entire test suite is deleted...

If the execution context is safe to delete, then you should be able to just remove `execution_context_class=UnforgivingExecutionContext,` from each of these and they should still be passing.

AlecRosenbaum (commented):
@syrusakbary Does it really make sense to block this change on unrelated test failures?

syrusakbary (approved):
Looks great, merging!

graphene/types/tests/test_schema.py
codebyaryan: I think that this part should be handled by graphql-core since it is that library that is providing the ExecutionContext, not us.

codebyaryan:
@syrusakbary I think you forgot to set `PYPI_API_TOKEN` or it is invalid.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,83 @@
...
+
+  coveralls_finish:
+    # check coverage increase/decrease
+    needs: tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Coveralls Finished
+        uses: AndreMiras/coveralls-python-action@develop
+
+  deploy:
```
jkimbo, line 64: Erm @codebyaryan @syrusakbary  I don't think you want to deploy to pypi on every pull request...

codebyaryan: Yeah, I had my doubt. Thanks for clarifying. Should this be shifted to another workflow instead that fires whenever a github version is created??

jkimbo: Yes that is usually the way to do it. 

I've already setup GitHub actions for the v2 branch so you can just copy that: https://github.com/graphql-python/graphene/pull/1349
Just need to change the Python versions that it runs.

codebyaryan: done with another PR!

codebyaryan:
@syrusakbary should a new release be made to make this feature available?

PR #1359, opened 2021-08-21
Fix actions

codebyaryan:
With the previous PR where we aimed to fix CI with github actions, the workflows had a few problems. This PR aims to fix that. 

What's been added/modified:
- [x] split the test workflow into three separate workflow (tests, coveralls, deploy).
- [x] the deploy workflow fires conditionally, and not with every push/ pull request.
- [x] add lint workflow and update deploy referencing the v2 branch (which was previously setup by @jkimbo )
- [x] fix the location of stale.yml
Also, python 3.8 is being primarily used, because it's the most stable version we have yet (over 3.9)

Also removes unused imports from test files.


codebyaryan:
@jkimbo @syrusakbary what do you suggest we do with the pre-commit hooks? Should they alter files when we commit (this might be a large change because every file is checked). Right now, the PR checks fail if there is any issue with the commit (like improper formatting or old python code).
We could alter the pre-commit hooks to just magically format and upgrade code on each commit, if that is preferred.

codebyaryan:
I'm also having trouble running pre-commit on my machine. black formatter fails because of Errno 13, as well as pyupgrade for some reason. I will try to fix this later on another machine,

syrusakbary:
Great work on the PR @codebyaryan.

The action should just check if the lint is correct, but don't update the code automatically (that should be done on the user side before pushing).
Could you run the linter locally and commit also the files with the formatted fixes?

Once that's done we should be good to merge!

codebyaryan:
Thanks for the reply! I will make this PR ready ASAP!

codebyaryan:
shouldn't the `continuous-integration/travis-ci` action be removed? (because we switched to github actions instead)

syrusakbary:
> shouldn't the continuous-integration/travis-ci action be removed? (because we switched to github actions instead)

Just fixed it!

syrusakbary (approved):

ulgens:
@syrusakbary I know it's too late but it would be nicer if we squashed this before merging 🙂

PR #1360, opened 2021-08-25
Make Decimal type return None if parsing an empty string

### LINKED ISSUE #927: Empty value for Decimal scalar
Is empty string considered an invalid value for Decimal scalar?

If so then how do we represent an empty value for a Decimal?

I've tried to fix this by parsing empty string to `None` but that results in an error. The error happens in `graphql.utils.is_valid_value` (see below).

```python
"""
This part of the code from `graphql.utils.is_valid_value` returns errors 
when None value is returned from Decimal scalar
"""
parse_result = type.parse_value(value)
    if parse_result is None:
        return [u'Expected type "{}", found {}.'.format(type, json.dumps(value))]
```

**UPDATE:**
**I dug my head out of the sand and realised that using `null` on the client is the obvious thing to do for clearing a value.**

abs25:
Would fix #927.

When using forms in the frontend, it's often useful for the "null" value of the form field to be an empty string, as HTML Input elements don't take null for a value. When converting that form value into a Graphene Decimal type, an empty string will currently cause a ConversionError. This PR makes empty strings a valid "null" value for Decimal.

### DIFF
--- a/graphene/types/decimal.py
+++ b/graphene/types/decimal.py
@@ -28,6 +28,8 @@ class Decimal(Scalar):
 
     @staticmethod
     def parse_value(value):
+        if value == "":
+            return None
         try:
             return _Decimal(value)
         except ValueError:
--- a/graphene/types/tests/test_decimal.py
+++ b/graphene/types/tests/test_decimal.py
@@ -6,7 +6,7 @@ from ..schema import Schema
 
 
 class Query(ObjectType):
-    decimal = Decimal(input=Decimal())
+    decimal = Decimal(input=Decimal(required=False))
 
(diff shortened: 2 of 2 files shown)


rzane:
Hey @abs25, I'm not a contributor, but I also encountered this issue. I think the issue here is that the original author of Graphene's Decimal type assumed that a `ValueError` would be raised in the case of a value that couldn't be parsed. However, that's not the case.

Rather than checking for a blank string,  I a better solution would be to change `except ValueError` to `except decimal.InvalidOperation`. This will handle scenarios where someone sends `"abc"` to the server as well.

Issue #1368, opened 2021-09-20
No longer able to catch unexpected errors after removal of UnforgivingExecutionContext


alexhafner:
* **What is the current behavior?**
Since the introduction of https://github.com/graphql-python/graphene/pull/1255 by @AlecRosenbaum it was possible to raise unexpected errors to the backend processing, as examplained in https://github.com/graphql-python/graphene/issues/902#issuecomment-672952043. This is critical so you get a decent traceback in the log and no error message for each record in the result set, ie

```json
{
  "errors": [
    {
      "message": "An unknown error occurred.",
      "locations": [
        {
          "line": 5,
          "column": 5
        }
      ],
      "path": [
        "testObject",
        0,
        "testField"
      ]
    },
    {
      "message": "An unknown error occurred.",
      "locations": [
        {
          "line": 5,
          "column": 5
        }
      ],
      "path": [
        "testObject",
        1,
        "testField"
      ]
    }
]
}
```

Changes in graphql-core started to let the UnforgivingExecutionContext tests fail, as recorded by @mweinelt in https://github.com/graphql-python/graphene/issues/1346 and noted by @weilu https://github.com/graphql-python/graphene/pull/1255#issuecomment-857877151 with a suggestion for a potential fix.

https://github.com/graphql-python/graphene/pull/1357 by @codebyaryan brought in welcome changes for query validation, but also removed UnforgivingExecutionContext in  https://github.com/graphql-python/graphene/pull/1357#issuecomment-902573578

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem** via
a github repo, https://repl.it or similar.

If we bring back the Test case for UnforgivingExecutionContext without specifying UnforgivingExecutionContext per this configuration https://github.com/alexhafner/graphene/commit/0aaa3fcfac05ae5243d0a65faa562334ae5f37e3, we get those tests failing because unexpected errors are no longer being raised up: https://gist.github.com/alexhafner/5215ef726b356eef6571efb969f6a3e7

* **What is the expected behavior?**

Be able to stop processing a graphql operation when an unexpected error occurs, return a top-level error message to the GraphQL user, and get a full stack trace in the backend to be able to act on the error message further. 

For example:

[...]

Issue #1346, opened 2021-07-03
TestUnforgivinExeuctionContext.test_unexpected_error fails


mweinelt:
**Note: for support questions, please use stackoverflow**. This repository's issues are reserved for feature requests and bug reports.

* **What is the current behavior?**

TestUnforgivinExeuctionContext.test_unexpected_error fails.

```
============================= test session starts ==============================
platform linux -- Python 3.9.5, pytest-6.2.4, py-1.10.0, pluggy-0.13.1
benchmark: 3.2.2 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: /build/source, configfile: tox.ini
plugins: asyncio-0.15.1, benchmark-3.2.2, mock-3.6.1, snapshottest-0.6.0
collected 389 items

examples/starwars/tests/test_query.py .............                      [  3%]
examples/starwars_relay/tests/test_connections.py .                      [  3%]
examples/starwars_relay/tests/test_mutation.py .                         [  3%]
examples/starwars_relay/tests/test_objectidentification.py ......        [  5%]
graphene/relay/tests/test_connection.py ...........                      [  8%]
graphene/relay/tests/test_connection_async.py .                          [  8%]
graphene/relay/tests/test_connection_query.py .....................      [ 13%]
graphene/relay/tests/test_global_id.py ....                              [ 14%]
graphene/relay/tests/test_mutation.py .........                          [ 17%]
graphene/relay/tests/test_mutation_async.py ..                           [ 17%]
graphene/relay/tests/test_node.py ..............                         [ 21%]
graphene/relay/tests/test_node_custom.py ...........                     [ 24%]
graphene/tests/issues/test_313.py .                                      [ 24%]
graphene/tests/issues/test_356.py .                                      [ 24%]
graphene/tests/issues/test_425.py .........                              [ 26%]
graphene/tests/issues/test_490.py .                                      [ 27%]
graphene/tests/issues/test_720.py .                                      [ 27%]
graphene/tests/issues/test_956.py .                                      [ 27%]
graphene/types/tests/test_argument.py ........                           [ 29%]
graphene/types/tests/test_base.py .....                                  [ 31%]
graphene/types/tests/test_base64.py .......                              [ 32%]
graphene/types/tests/test_datetime.py .............                      [ 36%]
graphene/types/tests/test_decimal.py ....                                [ 37%]
[...]

PR #1369, opened 2021-09-20
Re-create UnforgivingExecutionContext for  graphql-core 3.1.5+

alexhafner:
This pull request addresses the issue in #1368 where we want to be able to use UnforgivingExecutionContext to deal with unexpected errors. The tests for UnforgivingExecutionContext are re-instated as they were before https://github.com/graphql-python/graphene/commit/efc03533ae619e9716d6959d671cddaf8c6f7769#diff-dc90fef2a8de7ab6853d36140a9149eef3342af69e7d5f95e34b3fe2a9a8ebe1, and a slightly adapted of the fix proposed by @weilu https://github.com/graphql-python/graphene/pull/1255#issuecomment-857877151 is implemented for UnforgivingExecutionContext itself. The modified version differentiates between GraphQLError, which it leaves to be handled by graphql-core and other errors, which it treats as unexpected and raises with the original error. This should also fix the tests failing per #1346.

### DIFF
--- a/graphene/types/schema.py
+++ b/graphene/types/schema.py
@@ -11,6 +11,7 @@ from graphql import (
     print_schema,
     subscribe,
     validate,
+    ExecutionContext,
     ExecutionResult,
     GraphQLArgument,
     GraphQLBoolean,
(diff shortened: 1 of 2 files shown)


graphene/types/schema.py
```diff
@@ -54,6 +55,22 @@
...
+    Instead it re-raises the original error, except for
+    GraphQLError, which is handled by graphql-core
+    """
+
+    def handle_field_error(
+        self,
+        error: GraphQLError,
+        return_type,
+    ) -> None:
+        if type(error.original_error) is GraphQLError:
```
AlecRosenbaum, line 69: In our project, I had to change this to `isinstance(error.original_error, GraphQLError)`.

We have a few subclasses of `GraphQLError` to support transformations from different data types using `OurGraphQLError.from_other_error(...)`. This makes it easier to support `extension`'s within the errors.

PR #1375, opened 2021-09-27
Increased MAX_INT and MIN_INT to +(2^53 - 1) and -(2^53 - 1)

jackton1:
Seems the comment uses the [MAX_SAFE_INTEGER](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER#description) which should be 2^53 - 1 instead of 2^32 - 1.

https://github.com/graphql/graphql-spec/issues/73

This might be platform dependent so switching to use `sys.maxsize` might be an alternative.

@syrusakbary


erikwrede:
Using the BigInt Scalar is the preferred solution for this problem.

PR #1387, opened 2021-11-05
Chore: Refactor Multi Expression Code ✨

yezz123:
I just refactor some functions, mostly was simple like:
- Swap if & else branches.
- Merge Else If Into Elif.
- De Morgan Identity.

### DIFF
--- a/graphene/relay/node.py
+++ b/graphene/relay/node.py
@@ -18,11 +18,7 @@ def is_node(objecttype):
     if not issubclass(objecttype, ObjectType):
         return False
 
-    for i in objecttype._meta.interfaces:
-        if issubclass(i, Node):
-            return True
-
-    return False
+    return any(issubclass(i, Node) for i in objecttype._meta.interfaces)
 
 
 class GlobalID(Field):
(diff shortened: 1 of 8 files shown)


graphene/relay/node.py
```diff
@@ -92,9 +88,7 @@ def get_node_from_global_id(cls, info, global_id, only_type=None):
             _type, _id = cls.from_global_id(global_id)
         except Exception as e:
             raise Exception(
-                f'Unable to parse global ID "{global_id}". '
-                'Make sure it is a base64 encoded string in the format: "TypeName:id". '
-                f"Exception message: {str(e)}"
+                f'Unable to parse global ID "{global_id}". Make sure it is a base64 encoded string in the format: "TypeName:id". Exception message: {e}'
```
belkka, line 91: Probably original intention was to keep lines short enough. BTW, there is no line wrap on the left side panel of github diff view.

I would suggest to keep original string, except, without `str()` call inside f-string.

belkka (approved):
Great changes!

graphene/relay/node.py
yezz123: Fixed

syrusakbary (approved):

PR #1393, opened 2021-12-02
Define abstract Mutation.mutate() method to improve auto-completion in IDE

belkka:
Add default implementation of Mutation.mutate() that raises NotImplementedError.
This makes code more clear and also improves work of auto-completion tools (e. g. in IDE). They usually guess if user is overriding a class method and suggest to complete method name/arguments.

Not using python's abc.ABC because of possible metaclass conflicts.


belkka:
I will be glad to perform more similar refactoring in the project if current change will be considered reasonable.

graphene/types/mutation.py
```diff
@@ -106,9 +106,7 @@ def __init_subclass_with_meta__(
             else:
                 arguments = {}
         if not resolver:
-            mutate = getattr(cls, "mutate", None)
-            assert mutate, "All mutations must define a mutate method in it"
-            resolver = get_unbound_function(mutate)
+            resolver = get_unbound_function(cls.mutate)
```
belkka, line 109: Not sure that with current implementation it's really needed to call `get_unbound_function()` instead of just referring `cls.mutate`. Should I check it?

DarknessRdg: I think there is no need to use `cls.mutate` as argument to `get_unbound_function` , because it's already store in `mutate` variable. In this case the variable is just being used to validate on assert.

DarknessRdg:
I think your approach is really interesting due to IDE warning. The only difference between your implementations and previous one is the fact that your code only raises exception in runtime when the function `mutate` is called.

The older one would raise with class `__new__`  and the error is raises after the module is loaded.

DarknessRdg (commented):
I think your approach is really interesting due to IDE warning. The only difference between your implementations and previous one is the fact that your code only raises exception in runtime when the function `mutate` is called.

The older one would raise with class `__new__`  and the error is raises after the module is loaded.

For example, the following test fails with your code:
```py
def test_mutation_raises_exception_if_no_mutate():
    with raises(AssertionError) as excinfo:

        class MyMutation(Mutation):
            pass

    assert "All mutations must define a mutate method in it" == str(excinfo.value)
```

I think it's a break change if your code is merged. So, maybe there is another path to follow in order to validate if it's overwritten and also raises the the same exception.


Maybe a workaround would be to check if the method is implemented:

```py
def add_implemented_flag(func):

    def wrapper(*args, **kwargs):
        return func(*args, **kwargs)

    wrapper._is_implemented = False
    return wrapper



class Mutation(ObjectType):

    @classmethod
    @add_implemented_flag
    def mutate(cls, parent, info, **kwargs):
        raise NotImplementedError("All mutations must define a mutate method in it")

    @classmethod
    def __init_subclass_with_meta__(
        cls,
        interfaces=(),
        resolver=None,
        output=None,
        arguments=None,
        _meta=None,
        **options,
    ):
        ... 

        if not resolver:
            is_defined = getattr(cls.mutate, '_is_implemented', True)

            assert is_defined, "All mutations must define a mutate method in it"
            resolver = get_unbound_function(cls.mutate)
```

belkka:
Sorry for the late reply. Could have done it much earlier, but didn't feel that I have enough time to write down all my thoughts... and now I finally did :)

### 1. Is it a breaking change?

I do not feel it's a breaking change, because user code usually must not rely on the fact that library code raises an `AssertionError` (unless it's testing framework ofc). Imagine a library function raises an `AssertionError` for some combination of argument values.
```python
def foo(x):
    assert x in ("yes", "no"), "My developer did not expect other values"
```
  
Imagine that in new version of library we wanna update this function in a such way, that it can work with values `"on"`/`"off"`/`"maybe"/`None etc. Of course we could say that updated version does not pass an imaginary test
```python
def test_foo():
    with raises(AssertionError):
        foo("maybe")
```
But I do not think this is widely considered as "breaking backward compatibility". This may be not the best example, because for argument validation it's usually more appropriate to raise Type- or ValueError rather than AssertionError... and in that case the question becomes more tricky. It's valid for client code to catch `ValueError` using try..except and rely on the fact that library function raises a **documented** exception for some combination of inputs; so, indeed, sometimes it may be considered a breaking change to stop raising a ValueError/other exception. But semantics of `AssertionError` implies that it's never raised in a valid working program, so in general user code must never *expect* `AssertionError` to happen and, therefore, must not break if it doesn't happen. In this particular case, I also cannot imagine a working graphql app that somehow breaks when mutation class definition stops raising an AssertionError (after all, it is supposed that users DO define a `mutate` method).
 
### 2. Is current behavior (raise error at class definition time) so good?

@DarknessRdg while it may be possible to simplify your workaround (by doing a check like `cls.method is BaseClass.method` to avoid any decorators and additional attributes on methods), I still suggest that current behavior (raising an exception on class **definition**) is changed, and let me explain why. Conception we are talking here about is called an abstract class/abstract method. AFAIK there are two common approaches to define abstract methods in python — (1) `NotImplementedError` and (2) `abc.abstractmethod`. In this PR I've implemented the first one and, as you have noted, it is generally weaker, since it only shows an error at run time. The python built-in abc module allows to define abstract classes and could be a good alternative, but since it also relies on metaclasses, this approach may require that internal graphene metaclasses inherit from `abc.ABCMeta` (which may turn out to be a nice and clean solution btw, but it requires verification, as I did not try it). Note that abstract classes from `abc` and, for example, abstract classes in C++ do not actually forbid to **define** an abstract subclass. What both python abc.ABC and C++ do is they [forbid to instantiate](https://en.wikipedia.org/wiki/Abstract_type) abstract classes and subclasses unless a subclass overrides all abstract methods (so it becomes non-abstract). I think things are done this way for a reason and there are following practical considerations for that:

Consider if one wanna create a microframework on top of graphene or even simply customize the base `Mutation` class in their project? If I write

```python
class FancyBaseMutation(graphene.Mutation):
    fancy_attribute = None

    def cool_helper_method(self, ...):
        do_something(self.fancy_attribute)
        ...


class CreateMessage(FancyBaseMutation):
    fancy_attribute = FancyMessage
    
    @classmethod
    def mutate(cls, ...):
        self.cool_helper_method(...)
        ...
```

the first class definition now raises an error because it does not override an "abstract" method .mutate(). But it is not an error, because I supposed to use first class as base for real mutations. What is the easiest solution for this issue that comes into head? Just define a dummy mutate method in the base class:

```python
class FancyBaseMutation(graphene.Mutation):
    class_attribute = None
    
    @classmethod
    def mutate(cls, parent, ...):
        raise NotImplementedError("don't forget to override .mutate() method!")

    def cool_helper_method(self, ...):
        ...
```

Cool, graphene does not raise an AssertionError anymore + as a bonus IDE suggests you method name/arguments when you start type `def m...` it in subclasses. Though, this workaround kills the original idea of detecting errors at "import" time, now NotImplementedError is raised at run time. But instead of having a default .mutate() method in original `graphene.Mutation`, users are forced to define it by in a custom subclass (even if they don't really need it).

### 3. Is suggested behavior (raise error at run time) so bad?

The most "balanced" solution that allows earlier error detection while not breaking extensibility is to raise an error at class instantiation time (i. e. when Mutation instance is created and it's detected that it has no overridden .mutate() method), as it is done in abc. This can be achieved either by
- inheriting graphene metaclass(es) from `abc.ABCMeta` (as mentioned above) and decorating `.mutate()` as `abc.abstractmethod`, or
- re-implementing functionality of `abc.ABCMeta` in graphene classes (if first option does not work for some reason). Note, that `add_implemented_flag` decorator in your example is kind of reinvented `abc.abstractmethod`, so first option seems to be reasonable

Nevertheless, I don't feel that this "balanced" solution is worth it. Even if user (backend developer) does *not* wanna extend `graphene.Mutation` into custom base class --- what if they work on new API and just wanna start from choosing some names and thinking about overall structure, like


```python
class CreatePerson(Mutation):
   ...

class UpdatePerson(Mutation):
   ...
   
class AddFriend(Mutation):
   ...
   
class CreateMessage(Mutation):
   ...

class ForwardMessage(Mutation):
   ...
```

And what if after finishing with architecture they then wanna implement corresponding .mutate() methods one-by-one? They may wish to ignore non-implemented mutations for a while and be able to test other mutations as soon as they are completed; i. e. implement, check (maybe immediately write some tests), the move to the next non-implemented mutation. Currently "draft" mutation classes cannot be even defined like above, because graphene will raise an exception on module import. But even if we implement the "abc-like" behavior, this means that users will still be forced to add dummy mutation methods immediately in order to run some tests against implemented parts of API, like:

```
class CreatePerson(Mutation):
    @classmethod
    def mutate(cls, parent, *args, **kwargs):
         do_things()  # would be nice to test this ASAP, so I don't make same mistakes in UpdatePerson 
 
 class UpdatePerson(Mutation):
   @classmethod
    def mutate(cls, parent, *args, **kwargs):
         pass  # to do (adding this stub to test another mutation, otherwise graphene complains)

```

If they do this, then was the whole point of enforcing them to override .mutate methods? It's the same as having default "raise NotImplementedError" implementation, but worse, because user is forced to create these stubs manually

### TL;DR
While it is possible to benefit from IDE auto-completion and either keep the current behavior (raise error at class definition time) or implement abstract class behavior (raise error at class instantiation time), I believe that most user-friendly option is to raise `NotImplementedError` at run-tme (and also the most straightforward). I could try to implement the second option though.

PR #1399, opened 2022-01-06
Sets default_value for argument to Undefined instead of None

### LINKED ISSUE #1398: Undefined arguments always passed to resolvers
Contrarily to what's described [here](https://docs.graphene-python.org/en/latest/types/objecttypes/#graphql-argument-defaults): it looks like all the arguments with unset values are passed to the resolvers in graphene-3:

This is the query i am defining
```

    hello = graphene.String(required=True, name=graphene.String())

    def resolve_hello(parent, info, **kwargs):
        return str(kwargs)
```

which i submit as 
```
{
    hello
}
```

The result is : 
```
{
    "data": {
        "hello": "{'name': None}"
    }
}
```

The expected returned value is : 
```
{
    "data": {
        "hello": "{}"
    }
}
```

which is what we're getting with graphene-2.

My environment:
```
graphene==3.0
graphql-core==3.1.7
graphql-relay==3.1.0
graphql-server==3.0.0b4
```

stabacco:
fixes #1398 .
The tests' asserts are now matching the v2 ones, which I believe match the expected behaviour described here:

https://docs.graphene-python.org/en/latest/types/objecttypes/#graphql-argument-defaults

### DIFF
--- a/examples/starwars_relay/tests/snapshots/snap_test_objectidentification.py
+++ b/examples/starwars_relay/tests/snapshots/snap_test_objectidentification.py
@@ -7,31 +7,52 @@ from snapshottest import Snapshot
 
 snapshots = Snapshot()
 
-snapshots["test_correctly_fetches_id_name_rebels 1"] = {
-    "data": {
-        "rebels": {"id": "RmFjdGlvbjox", "name": "Alliance to Restore the Republic"}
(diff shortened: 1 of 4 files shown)


lovetoburnswhen:
Gah such a small change and yet this bug continues to make graphene unusable

stabacco:
Closing, was fixed in #1412

PR #1402, opened 2022-01-13
Highlight .get in backticks

RJ722:
When I first read through the documentation twice, it took me two tries and looking very hard to find out the difference b/w the two. The background highlight using backticks would be helpful in this case.


stabacco (approved):

ruohola (approved):

erikwrede (approved):
Looks good, thank you!

Issue #1394, opened 2021-12-07
Required/NonNull parameters do not work


Tomms93:
Required=True and NonNull parameters do not work in Query.
I have tried 3 following ways:
    hello2 = String(name=String(required=True))
    hello2 = String(name=NonNull(String))
    hello2 = String(name=Argument(String, required=True))

with following resolver:
    def resolve_hello2(root, info, name):
        return f"Hello {name}!"

And in any of these ways, when I do not provide name, I do not get any error:

![image](https://user-images.githubusercontent.com/95684343/145007133-42e17216-d3e5-4f4e-85c1-abda214762a8.png)

Although playground seems to recognize that 'hello2' need some parameters, because it is underlined, where 'hello' is not (it does not have any mandatory parameters):

![image](https://user-images.githubusercontent.com/95684343/145007353-ba14cb97-e62a-41b7-930a-512ac1a5274d.png)

Graphene version: 
graphene             3.0
graphql-core        3.1.6
graphql-relay       3.1.0

Running API using starlette_graphene3:
starlette-graphene3 0.5.1

Also tested running API using starlette.graphql  - the same behavior.

Python: 3.8.10

stabacco:
I am seeing the same behavior.

rarenatoe:
Same behavior here. Shameless bump.

tcleonard:
was solved by PR #6

flipbit03:
This is not solved, and is happening to me as well. Can we please reopen this bug? I have the same behavior as @Tomms93, @stabacco and @rarenatoe 

Relevant packages in my setup:

```
graphene==3.0
graphene-pydantic==0.3.0
graphene-sqlalchemy @ git+https://github.com/graphql-python/graphene-sqlalchemy.git@0820da77d94d947e35325ba40eea96462bc890a7
graphene-sqlalchemy-filter @ git+https://github.com/AsimovBio/graphene-sqlalchemy-filter.git@0ec819cef3bdae6f48282ae9a8b8e2e4c22ca830
graphql-core==3.1.7
graphql-relay==3.1.5
starlette-graphene3==0.5.1
```

I can run any kind of resolver/mutation without required parameters. They just appear as `None` in the mutation/resolver function. This is potentially a showstopper bug for me.

tcleonard:
This was released in graphene >=3.1 so it's normal you still get it if you are using versions below that

PR #1412, opened 2022-03-20
fix: default value for argument should be Undefined (Issue #1394)

tcleonard:


stabacco:
A similar fix was provided in #1399 which has been sitting there for a long time unfortunately. Not sure who's supposed to have a look at these but it'd be great if someone could. 
This issue is currently holding us back from being able to upgrade all our gql stack.

Issue #1413, opened 2022-03-20
Argument with wrong value type do not raise an error


tcleonard:
Currently, setting the value of an input with the wrong type is not raising an error.
For example if the input as the type `UUID` and that you provide the value `2` it doesn't fail.

There are various cases like that (see tests in the pull request for an extensive list).

This is due to the fact that now no value is represented by `Undefined` and not `None` in graphql-core.

DarknessRdg:
Hi, @tcleonard !

I'm just commenting the link to your PR:  https://github.com/graphql-python/graphene/pull/1414

erikwrede:
Merged!

anusha-bylla:
“Hi, I am new to Qt Creator. I successfully ran QGroundControl on Windows, but now I am trying to run QGroundControl for Android in Qt Creator, and I am encountering the following error:”
can you please help me to solve this error…

![Screenshot (7)](https://github.com/user-attachments/assets/23826e33-27eb-44b4-826a-2fefdd3e0443)
![Screenshot (8)](https://github.com/user-attachments/assets/b9a77d1e-db6f-43a6-bc25-d831a118ed11)

PR #1414, opened 2022-03-20
Issue #1413 fix invalid input type

tcleonard:
The failure in the CI exists on master as well.
You can find the fix on the PR #1412

### DIFF
--- a/graphene/types/json.py
+++ b/graphene/types/json.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import
 
 import json
 
+from graphql import Undefined
 from graphql.language.ast import StringValueNode
 
 from .scalars import Scalar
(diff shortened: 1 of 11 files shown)


graphene/types/json.py
```diff
@@ -22,7 +23,11 @@ def serialize(dt):
     @staticmethod
     def parse_literal(node):
         if isinstance(node, StringValueNode):
-            return json.loads(node.value)
+            try:
+                return json.loads(node.value)
+            except Exception:
+                raise ValueError("badly formed JSONString")
```
erikwrede, line 29: Wouldn't it make sense to add the underlying exception message here? Error handling will probably only add"badly formed JSONString" to the query response - I always find it helpful if the error message assists in finding my JSON typo / malformed part when writing queries.

erikwrede (commented):
+1, looks good to me!

erikwrede (approved):
LGTM!

PR #1426, opened 2022-06-15
Union types documentation

ramonwenger:
Adds some documentation and examples to help those struggling with the type resolution of Unions (like I myself was just recently).

As this is my first real contribution to this project I am glad for any pointers or improvements to get my PR accepted without an unnecessary delay. Thanks!

### DIFF
--- a/docs/types/unions.rst
+++ b/docs/types/unions.rst
@@ -61,3 +61,63 @@ The above types have the following representation in a schema:
 
     union SearchResult = Human | Droid | Starship
 
+Resolving Types
+~~~~~~~~~~~~~~~
+
+When defining Unions, as with Interfaces, we need to tell the schema how to resolve the type of a returned object.
+
+This can be achieved by:
+
+- defining an `is_type_of`-method on each `ObjectType`
+- defining the attribute `possible_types` on the Meta class
+- defining a `resolve_type` on the Union
+
+Examples:
+^^^^^^^^^
+
+An example with `is_type_of` and `Meta.possible_types`:
+
+.. code:: python
+
+    class one_object:
+        one = 'one'
+
+    class two_object:
+        two = 'two'
+
+    class One(ObjectType):
+        one = String()
+
+        @classmethod
+        def is_type_of(cls, root, info):
+            return isinstance(root, one_object)
+
+    class Two(ObjectType):
+        class Meta:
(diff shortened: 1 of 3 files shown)


erikwrede:
What version of Graphene are you on? This works for me (3.0):


```
from fastapi import FastAPI
import graphene
from starlette_graphene3 import GraphQLApp, make_playground_handler


class A(graphene.ObjectType):
    a = graphene.String()

class B(graphene.ObjectType):
    b = graphene.String()

#BCD = graphene.Union([A,B])


class BCD(graphene.Union):
    class Meta:
        types=[A,B]
        name="ABC"

def create_app() -> FastAPI:
    app = FastAPI()

    # graphql endpoint

        abc = graphene.Field(BCD)

        @classmethod
        def resolve_abc(cls,a,b):
            return A(a="Hi")

    app.mount("/graphql", GraphQLApp(schema, on_get=make_playground_handler()))

    return app


app = create_app()

```

erikwrede:
And it should be working just like that due to this method:

https://github.com/graphql-python/graphene/blob/72c2fd5ec38af595519b2ab334696397760d2d81/graphene/types/union.py#L74

ramonwenger:
@erikwrede Agreed, and yes, there are different use cases. I have run into problems with this using the Django integration with the example as provided in this repository, after I've started creating some custom types for use in an Union.

I myself would have been glad to have a clearer distinction between using an `ObjectType` and something "taking the form" of one, with the implications of the latter. Also, I was not quite sure if it's idiomatic or "the right thing" to instance ObjectTypes like you do in your example, or if the other way with plain object is preferred. As I'm now aware, the way to go would probably to just use an `ObjectType`, but that could also be made a bit clearer in the documentation, in my opinion. 

Would you agree to the above, @erikwrede?  Or did I make some false assumption somewhere?

erikwrede:
Well, docs need a general overhaul so issues like you described with the object types have clear solution. 
I like your documentation on unions, but I think it should include the case I mentioned since it is the simplest of all cases not requiring any classmethods or additional fields. If you want to add that, I'll happily merge this PR. 
If you want to get involved with more docs overhaul, please contact me on slack so we can discuss things further 🙂 

Best
Erik

erikwrede (changes requested):
See comment

PR #1441, opened 2022-08-13
Fix Test Failure due to #1304

erikwrede:
`assert bar_graphql_type.interfaces == [foo_graphql_type]` failed only on tox, because .interfaces was a tuple instead of a list. Error didn't occur using just pytest. Fixed by explicitly 
converting both to list.


erikwrede:
Closed; Solved by commit to master

PR #1459, opened 2022-09-05
Vendor `DataLoader` from `aiodataloader` and move `get_event_loop()` out of `__init__` function.

flipbit03:
DataLoaders are trying to get the asyncio's event loop too early (on `__init__`) and this is generating problems like [making pytest + async unusable](https://github.com/syrusakbary/aiodataloader/issues/13). Since there is no event loop at that moment, the first DataLoader instantiated is actually creating a new event loop. This generates the infamous "Future tied to a different loop" problem many people are facing.

In this PR, we are:
- Vendoring `aiodataloader`'s `DataLoader` class into the `utils/` module
- Adding a fix that moves `get_event_loop` out of the class' initializer and into a property which only gets called when actually needed, offsetting the chicken-and-egg problem we are dealing with.

The base for this work is dataloader.py, which is MIT Licensed code from Syrus Akbary


setup.cfg
```diff
@@ -2,6 +2,10 @@
 exclude = setup.py,docs/*,*/examples/*,graphene/pyutils/*,tests
 max-line-length = 120
 
+# This is a specific ignore for Black+Flake8
+# source: https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#id1
+extend-ignore = E203
```
flipbit03, line 7: without this line, flake8 was constantly clashing with black over array slice notation whitespacing rules.

erikwrede:
Seems like `test_dataloader` is never run, explaining the decrease in coverage

erikwrede (approved):

PR #1461, opened 2022-09-29
hashable Enum

bkad:
It would be convenient for `graphene.Enum` to be hashable the same way `enum.Enum` is so that it can be used as a key in a dict or a member of a set.

### DIFF
--- a/graphene/types/tests/test_enum.py
+++ b/graphene/types/tests/test_enum.py
@@ -518,3 +518,28 @@ def test_mutation_enum_input_type():
     assert result.data == {"createPaint": {"color": "RED"}}
 
     assert color_input_value == RGB.RED
+
+
+def test_hashable_enum():
+    class RGB(Enum):
+        """Available colors"""
+
+        RED = 1
+        GREEN = 2
+        BLUE = 3
+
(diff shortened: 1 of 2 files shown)


graphene/types/tests/test_enum.py
```diff
@@ -518,3 +518,18 @@ class Query(ObjectType):
...
+
+
+def test_hashable_enum():
+    class RGB(Enum):
+        """Available colors"""
+
+        RED = 1
+        GREEN = 2
+        BLUE = 3
+
```
erikwrede, line 530: This should also include the instance-creation pattern for enums, to ensure completeness:
https://docs.graphene-python.org/en/latest/types/enums/#definition

```python
Episode = graphene.Enum('Episode', [('NEWHOPE', 4), ('EMPIRE', 5), ('JEDI', 6)])
```

erikwrede (changes requested):
Thanks for the PR! Small comment, other than that it is good to merge! 🚀

graphene/types/tests/test_enum.py
bkad: Thanks! I added a test in my amended commit.

erikwrede (approved):

erikwrede:
Thank you!

PR #1466, opened 2022-10-14
feat: added object_type_name_prefix

superlevure:


erikwrede:
Thanks for the PR, this looks helpful!
Will test it out locally tomorrow.

erikwrede:
Quick comment which is not a full review: This also prefixes the root type names. I'm wondering if this is really desirable, as `Query`, `Mutation` and `Subscription` types are basically the standard/usual names.  The prefixes will cause Docs to deviate from the usual pattern. Maybe a way to edit this behavior would be useful.

image

erikwrede (commented):
After testing, I can conclude that this is indeed useful to have. I have small comments (including my comment that is not part of the review):

- Is there a reason we are only prefixing `ObjectTypes`, but excluding things like `Enums`, `I terfaces` and `Unions`? Especially in a Microservice scenario like the one you mentioned these would most likely also be specific to the Microservice. The parameter name sort of specifies that the prefix is only applicable to `ObjectTypes`, but I could imagine this restriction to be confusing for other devs.

superlevure:
Hi @erikwrede 

Thanks for the review! I'm glad you think that it could be useful to Graphene. 

> Quick comment which is not a full review: This also prefixes the root type names. I'm wondering if this is really desirable, as `Query`, `Mutation` and `Subscription` types are basically the standard/usual names. The prefixes will cause Docs to deviate from the usual pattern. Maybe a way to edit this behavior would be useful.
> 
> image
> 
> My intuition is that excluding root types from prefixing actually helps with schema stitching and general readability of a schema. Happy to have my mind changed on this one though, I'm open to discuss it if you have any compelling reasons! 🙂

Absolutely, I was not sure about this one but I agree. I'll push a fix


> Is there a reason we are only prefixing ObjectTypes, but excluding things like Enums, I terfaces and Unions? Especially in a Microservice scenario like the one you mentioned these would most likely also be specific to the Microservice. The parameter name sort of specifies that the prefix is only applicable to ObjectTypes, but I could imagine this restriction to be confusing for other devs.

I missed that indeed, we should apply the prefix to those too.

> Can we add a test case for the Field("Typename-As-import-String") notation? Just want to make sure future changes to that syntax aren't broken by prefixes. This might be the case if we ever decide to allow Field("Typename") using the schema as a registry instead of import strings.

Sure thing!

superlevure:
I've updated the PR to address your comments @erikwrede:

- Root type names are left untouched
- Prefix is now added to the name of:
  - ObjectType
  - InputType
  - Interface
  - Union
  - Enum
  - Query / Mutation / Subscription
- Added [a test ](https://github.com/superlevure/graphene/blob/feat/add-optional-prefix/graphene/types/tests/test_schema.py#L50)for the `Field("Typename-As-import-String")` notation
- Updated documentation

Let me know what you think :)

graphene/types/schema.py
```diff
@@ -164,8 +174,7 @@ def create_scalar(graphene_type):
             parse_literal=getattr(graphene_type, "parse_literal", None),
         )
 
-    @staticmethod
```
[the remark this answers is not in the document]
tcleonard: not sure it is needed for it to be a static method... but I guess it also does not need `self` so it could totally be static. And there is no real downside of having it as a static method. (also that needs to be consistent with `create_scalar`.
I would revert this change in my opinion...

graphene/types/schema.py
```diff
@@ -391,6 +411,23 @@ def resolve_type(self, resolve_type_func, type_name, root, info, _type):
         return_type = self[type_name]
         return default_type_resolver(root, info, return_type)
 
+    def add_prefix_to_type_name(self, name):
+        if self.type_name_prefix:
+            return self.type_name_prefix[0].upper() + self.type_name_prefix[1:] + name
```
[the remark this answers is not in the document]
tcleonard:
```suggestion
            return self.type_name_prefix.capitalize() + name
```

graphene/types/schema.py
```diff
@@ -164,8 +174,7 @@ def create_scalar(graphene_type):
             parse_literal=getattr(graphene_type, "parse_literal", None),
```
tcleonard, line 174: So apparently I can't leave a comment on the line I actually want to leave the comment so I'll put it here... 
Shouldn't the `create_scalar` method above also get the same modif of the name?

graphene/types/schema.py
```diff
@@ -357,7 +371,13 @@ def create_fields_for_type(self, graphene_type, is_input_type=False):
...
                     description=field.description,
                 )
-            field_name = field.name or self.get_name(name)
+            if field.name:
+                field_name = field.name
+            else:
+                if graphene_type._meta.name in self.root_type_names:
+                    field_name = self.add_prefix_to_field_name(name)
+                else:
+                    field_name = self.get_name(name)
```
tcleonard, line 380: I'm not completely sure but I think that even if the name comes from `field.name` we want to add the prefix... would be good to add a unit test for this.

```suggestion
            field_name = field.name or self.get_name(name)
            if graphene_type._meta.name in self.root_type_names:
                field_name = self.add_prefix_to_field_name(name)
```

graphene/types/schema.py
```diff
@@ -391,6 +411,23 @@ def resolve_type(self, resolve_type_func, type_name, root, info, _type):
...
+        if self.type_name_prefix:
+            if self.auto_camelcase:
+                return self.get_name(
+                    self.type_name_prefix[0].lower()
+                    + self.type_name_prefix[1:]
+                    + "_"
+                    + name
+                )
+            return self.get_name(self.type_name_prefix + name)
+        return self.get_name(name)
```
tcleonard, line 429: I don't think there is a need to deal with the `auto_camelcase` case here as it is already dealt with in the `get_name`.

```suggestion
    def add_prefix_to_field_name(self, name):
        if self.type_name_prefix:
           return self.get_name(
                self.type_name_prefix[0].lower()
                + self.type_name_prefix[1:]
                + "_"
                + name
            )
        return self.get_name(name)
```

also note that in your current implementation where you don't `lower` the first character in the case we are not in auto camel case means the output is not exactly what you give in the doc:
```
    type Query {
        myPrefixInner: MyPrefixMyType
    }
```
indeed that would be `MyPrefixInner`...

graphene/types/schema.py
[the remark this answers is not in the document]
superlevure: I do need `self` to add the prefix actually: https://github.com/graphql-python/graphene/pull/1466/files#diff-f066ce1b84329fbfb7cd0e5d5027a51f68e14677754aab172fbc91dea42c519dR205

graphene/types/schema.py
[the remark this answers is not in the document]
superlevure: I did that at first actually, but:

image

graphene/types/schema.py
superlevure: Looking again at the code I see that we bypass `auto_camelcase` when the field name is explicitly defined, shouldn't we expect the same for `type_name_prefix` ? 

I'm not sure neither to be honest

graphene/types/schema.py
superlevure: Absolutely, thanks for the catch. Fixed

graphene/types/schema.py
superlevure: > indeed that would be MyPrefixInner...

By default `auto_camelcase=True` so the query name would be `myPrefixInner` indeed isn't it?

graphene/types/schema.py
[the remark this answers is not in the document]
superlevure: (I've removed the `@static_method` for `create_scalar` too)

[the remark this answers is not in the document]
tcleonard: makes sense 👍

graphene/types/schema.py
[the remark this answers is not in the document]
tcleonard: mmmm that's a fair point! (that also means that if we have auto camel case on and we have `my_fieldIsGreat = Field(...)` it would become `myFieldisgreat` because it does use `.capitalize()`... should we change that @erikwrede in your opinion?)

superlevure:
I've pushed some changes to add the ability to override `type_name_prefix` for specific types @tcleonard @erikwrede 

I realized that this is needed when defining external GraphQL Types in a federated Schema to avoid ending up with `ServiceAServiceBTypeC`

Let me know what you think :)

graphene/types/schema.py
tcleonard: yeah I saw that we bypass the auto camel case but I am not sure it makes sense to remove the prefixing... 🤷‍♂️

graphene/types/schema.py
tcleonard: yes but my point was if `auto_camelcase` is:
- `True` ->  `myPrefixInner`
- `False`  -> `MyPrefixInner`
which doesn't seem very intuitive... no?

superlevure: I'm confused, what would be more intuitive in your view? 

Because 

* `auto_camelcase=True` -> `myPrefixInner` (`camelCase`)
* `auto_camelcase=False` -> we don't modify the prefix -> `MyPrefixInner` (`PascalCase`) 

Feels the most natural to me 🤔

graphene/types/schema.py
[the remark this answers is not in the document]
erikwrede: This makes it a breaking change with potential implications for downstream libs. Is there any way to avoid that?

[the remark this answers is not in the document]
erikwrede: I'm thinking something like an optional arg would easily prevent that. We could add a note to overhaul the syntax in the next breaking release. But I'm personally against pushing even small breaking changes in patches/minors. Open to discuss this though!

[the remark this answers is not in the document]
superlevure: That's totally fair, I've updated the code let me know

graphene/types/schema.py
superlevure: Gentle ping on that one @tcleonard

graphene/types/schema.py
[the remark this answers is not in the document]
erikwrede: Looks good now, thanks!

erikwrede:
@superlevure will test this in combination with some graphene federation 2.0 tests over the weekend, because that's probably the  main use case for this feature. Going to make sure we haven't created any incompatibilities by accident.

superlevure:
> @superlevure will test this in combination with some graphene federation v2.0 support reviews ([graphql-python/graphene-federation#4](https://github.com/graphql-python/graphene-federation/pull/4) ) over the weekend, because that's probably the main use case for this feature. Going to make sure we haven't created any incompatibilities by accident.

Thank you! 🙏 

Let me know if I can do anything to help

erikwrede:
@superlevure if you have experience using federation, then trying out the PR I mentioned would definitely be very useful.

graphene/types/schema.py
```diff
@@ -391,6 +412,26 @@ def resolve_type(self, resolve_type_func, type_name, root, info, _type):
...
         return default_type_resolver(root, info, return_type)
 
+    def get_type_name(self, graphene_type):
+        return get_type_name(graphene_type, self.type_name_prefix)
+
+    def get_field_name(self, graphene_type, name):
+        if graphene_type._meta.name in self.root_type_names:
+            # We only add the prefix to the root types and types defined prefixes take precedence
+            # over schema defined prefix.
+            type_name_prefix = (
```
erikwrede, line 422: Both methods should either be defined here or in utils.

erikwrede (commented):
As promised, I played around with this feature and `graphene-federation`. The basics (`@key`) are working if you manually disable the prefix for those types. 
More advanced directives such as `inaccessible` cause problems since their resolving relies on getting the graphql-core type from the schema by `schema.graphql_schema.get_type(entity._meta.name)`. Since `_meta.name` doesn't include the prefix, `get_type` returns `None`, breaking the schema generation.
Other extensions of graphene probably use this functionality as well.

Should we try to attach the prefix to `schema.get_type` in case the first search without the attached prefix fails, or should we let the extensions like `-federation` handle this?

graphene/types/schema.py
superlevure: Hey @erikwrede sorry I haven't gotten back to you yet, I'm busy at work at the moment but I haven't forgotten about this PR

erikwrede: @superlevure great to hear from u! no worries, feel free to tag me once you get to it 🙂

erikwrede:
Hey @superlevure any updates? Would love to get this merged soon 😊

superlevure:
> Hey @superlevure any updates? Would love to get this merged soon 😊

Hey @erikwrede, sorry I haven't had the chance to get back at it yet; it is still on my radar though

PR #1467, opened 2022-10-18
AttributeError: Type "get_type" not found in the Schema

allen-munsch:
```
    relay.Node.get_node_from_global_id(info, g_id) for g_id in global_id
  File "/home/jm/pycharm_projects/x/venv_3_10_7/lib/python3.10/site-packages/graphene/relay/node.py", line 98, in get_node_from_global_id
    graphene_type = info.schema.get_type(_type)
  File "/home/jm/pycharm_projects/xvenv_3_10_7/lib/python3.10/site-packages/graphene/types/schema.py", line 451, in __getattr__
    if isinstance(_type, GrapheneGraphQLType):
AttributeError: Type "get_type" not found in the Schema
```

graphene==3.1.1
python 3.10.7

### DIFF
--- a/graphene/types/schema.py
+++ b/graphene/types/schema.py
@@ -452,6 +452,9 @@ class Schema:
             return _type.graphene_type
         return _type
 
+    def get_type(self, name):
+        return self.graphql_schema.type_map.get(name)
+    
     def lazy(self, _type):
         return lambda: self.get_type(_type)
 
--- a/graphene/types/schema.py
+++ b/graphene/types/schema.py
@@ -453,8 +453,12 @@ class Schema:
         return _type
 
     def get_type(self, name):
-        return self.graphql_schema.type_map.get(name)
-    
+        '''
+        This will actually return a `GraphQLObjectType` of type `GrapheneObjectType` from graphql-core, not a graphene ObjectType.
+            These two are fundamentally different classes. That's the reason why your test is failing. You'd need to return the Graphene type here.
+        '''
+        return self.graphql_schema.type_map.get(name).graphene_type
+
     def lazy(self, _type):
         return lambda: self.get_type(_type)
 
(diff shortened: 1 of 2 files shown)


graphene/types/schema.py
```diff
@@ -452,6 +452,9 @@ def __getattr__(self, type_name):
             return _type.graphene_type
         return _type
 
+    def get_type(self, name):
+        return self.graphql_schema.type_map.get(name)
```
erikwrede, line 456: This will actually return a `GraphQLObjectType` of type `GrapheneObjectType` from `graphql-core`, not a graphene `ObjectType`. These two are fundamentally different classes. That's the reason why your test is failing. You'd need to return the Graphene type here.

erikwrede (changes requested):
Great catch!
Sadly the fix doesn't seem to work (see my attached comment)
Can you add a unit test for the methods that were experiencing the error so we can avoid future regressions?

graphene/types/schema.py
allen-munsch: @erikwrede Thanks, I tried to update the change based on your feedback. How's that look?

allen-munsch: although i'm not sure entirely, as now i get:

```
  File "/home/circleci/project/venv_3_9_10/lib/python3.9/site-packages/graphene/relay/node.py", line 116, in get_node_from_global_id
    graphene_type = graphene_type.graphene_type
```

erikwrede:
I'm wondering if this fixes the error you got with relay. Can you post a minimal example to reproduce the error please? It might not be caused by this after all.

erikwrede:
Due to a lack of reproduction I'm closing this for now. If this comes up again, feel free to post an issue with a reproduction example and we can reopen this!

allen-munsch:
@erikwrede I think this might have been the reproduction:

```
relay.Node.get_node_from_global_id(None, some_global_id)
```
Which, is probably more of an issue of how we were creating a `BaseMutation` parent class

Because of this, `info` was `None` and that might have led to the confusion here.

PR #1471, opened 2022-11-10
Changed InputObjectType's default builder-from-dict argument to be `Undefined` instead of `None` (or else we can't catch "null" inputs in InputObjectTypes, as both would coerce to `None`)

flipbit03:
This pull request aims to fix a shortcoming in the InputObjectTypes' implementation that makes it unable to differentiate an incoming `undefined` value from an intentional `null`.

The problem lies in the way the InputObjectType is constructed from the incoming parameter dictionary:

```py
class InputObjectTypeContainer(dict, BaseType):
    class Meta:
        abstract = True

        dict.__init__(self, *args, **kwargs)
        for key in self._meta.fields:
            setattr(self, key, self.get(key, None)) # <-- Ambiguous default value

    def __init_subclass__(cls, *args, **kwargs):
        pass
```

With `None` as that `self.get`'s default value, we are unable to differentiate between someone passing an *intentional* `null` (`None`) in an optional field. This PR refactors that default to point to `Undefined`, aligning closely with the GraphQL Spec.

This change is a potential breaking change for resolvers, so I've implemented a function called **set_input_object_type_default_value** that allows one to set the desired default value for InputObjectTypes. Calling `set_input_object_type_default_value(graphql.Undefined)` activates the new behavior in your codebase. This function should be called before your InputObjectTypes are defined.

The final container type then became:

```py
_INPUT_OBJECT_TYPE_DEFAULT_VALUE = None

def set_input_object_type_default_value(default_value):
    global _INPUT_OBJECT_TYPE_DEFAULT_VALUE
    _INPUT_OBJECT_TYPE_DEFAULT_VALUE = default_value

class InputObjectTypeContainer(dict, BaseType):  # type: ignore
    ...
        dict.__init__(self, *args, **kwargs)
        for key in self._meta.fields:
            setattr(self, key, self.get(key, _INPUT_OBJECT_TYPE_DEFAULT_VALUE))
    ...
```

### DIFF
--- a/graphene/types/inputobjecttype.py
+++ b/graphene/types/inputobjecttype.py
@@ -14,6 +14,14 @@ class InputObjectTypeOptions(BaseOptions):
     container = None  # type: InputObjectTypeContainer
 
 
+_INPUT_OBJECT_TYPE_DEFAULT_VALUE = None
+
+
+def set_input_object_type_default_value(default_value):
+    global _INPUT_OBJECT_TYPE_DEFAULT_VALUE
+    _INPUT_OBJECT_TYPE_DEFAULT_VALUE = default_value
+
+
 class InputObjectTypeContainer(dict, BaseType):  # type: ignore
     class Meta:
         abstract = True
@@ -21,7 +29,7 @@ class InputObjectTypeContainer(dict, BaseType):  # type: ignore
     def __init__(self, *args, **kwargs):
         dict.__init__(self, *args, **kwargs)
         for key in self._meta.fields:
-            setattr(self, key, self.get(key, None))
+            setattr(self, key, self.get(key, _INPUT_OBJECT_TYPE_DEFAULT_VALUE))
 
     def __init_subclass__(cls, *args, **kwargs):
         pass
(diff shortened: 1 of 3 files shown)


flipbit03:
@erikwrede as requested, I have provided a specific test for the new functionality around `InputObjectTypes`, so that we can have its default type be changed dynamically to whatever the user might want (a good suggestion is `Undefined` so that we can easily keep track of inputs that were _not_ provided in the Input). This behavior is `opt-in` so as not to break existing codebases.

IslamMesha (dismissed):

erikwrede:
Since no one had any adverse opinions here or on slack, I'd say it's time to get this change out soon! 

For consistency, we should also override the Argument default values with the global settings so that missing arguments are also marked `Undefined` instead of `None`.

flipbit03:
@erikwrede I believe this was already fixed in [here](https://github.com/graphql-python/graphene/commit/19ebf08339263f0446235ddffbbab0e6d3f11699) (#1394)

Issue #1446, opened 2022-08-18
Implement deprecation directive for field args, directive args and input fields


flbraun:
The GraphQL spec finally allows using the `@deprecated` directive on field args, directive args and input fields (see graphql/graphql-spec#805)

It would be nice to support this in Graphene using the established `deprecated_reason='....'` kwarg to the respective constructors.

erikwrede:
Sounds like a great addition!

TODOS for this:
- [ ] Is this already supported in graphql-core?
- [ ] Add deprecation reason field to Meta Options Classes and [Input]Field logic 
- [ ] Check that required NonNull input fields are not deprecatable (see spec)

If someone is willing to work on a PR I'd be happy to assist and review!

vhutov:
@erikwrede Hi Erik. Is this issue still relevant? I'd like to give it a try

erikwrede:
@vhutov Yes this is still relevant. Sounds great! Feel free to tag me when you need a review 🙂

erikwrede:
Will be  released in `2.1.2`, thanks to @vhutov! 🎉

PR #1472, opened 2022-11-12
Complete deprecated fields and arguments support

vhutov:
**Implementing enhancement from #1446**
In this PR I ensure that `Argument`, `Field` and `InputField` support **deprecation** functionality.

TODOs from the issue:

- [x]  Is this already supported in graphql-core?
- [ ] Add deprecation reason field to Meta Options Classes and [Input]Field logic 
       Skipped this one, as meta is not populated to the Mounted object.
- [x] Check that required NonNull input fields are not deprecatable (see spec)

Notes:
1. `Field` and `InputField` already had this field, whereas `Argument` didn't.
2. When constructing schema, some objects were already passing deprecation reason. I added the missing parts: `InputField` and `Argument`. `Enum` and `Field` were compliant.
3. When mounting `InputObject` or `Enum` **implicitly**, meta fields are not populated to the `Mounted` object.

### DIFF
--- a/graphene/types/argument.py
+++ b/graphene/types/argument.py
@@ -31,18 +31,22 @@ class Argument(MountedType):
         type (class for a graphene.UnmountedType): must be a class (not an instance) of an
             unmounted graphene type (ex. scalar or object) which is used for the type of this
             argument in the GraphQL schema.
-        required (bool): indicates this argument as not null in the graphql schema. Same behavior
+        required (optional, bool): indicates this argument as not null in the graphql schema. Same behavior
             as graphene.NonNull. Default False.
-        name (str): the name of the GraphQL argument. Defaults to parameter name.
-        description (str): the description of the GraphQL argument in the schema.
-        default_value (Any): The value to be provided if the user does not set this argument in
+        name (optional, str): the name of the GraphQL argument. Defaults to parameter name.
+        description (optional, str): the description of the GraphQL argument in the schema.
+        default_value (optional, Any): The value to be provided if the user does not set this argument in
             the operation.
+        deprecation_reason (optional, str): Setting this value indicates that the argument is
+            depreciated and may provide instruction or reason on how for clients to proceed. Cannot be
+            set if the argument is required (see spec).
     """
 
     def __init__(
         self,
         type_,
         default_value=Undefined,
+        deprecation_reason=None,
         description=None,
         name=None,
         required=False,
@@ -51,12 +55,16 @@ class Argument(MountedType):
         super(Argument, self).__init__(_creation_counter=_creation_counter)
 
         if required:
+            assert (
+                deprecation_reason is None
+            ), f"Argument {name} is required, cannot deprecate it."
             type_ = NonNull(type_)
 
         self.name = name
         self._type = type_
         self.default_value = default_value
         self.description = description
+        self.deprecation_reason = deprecation_reason
 
     @property
     def type(self):
@@ -68,6 +76,7 @@ class Argument(MountedType):
             and self.type == other.type
             and self.default_value == other.default_value
             and self.description == other.description
+            and self.deprecation_reason == other.deprecation_reason
         )
 
 
--- a/graphene/types/inputfield.py
+++ b/graphene/types/inputfield.py
@@ -55,11 +55,14 @@ class InputField(MountedType):
         description=None,
         required=False,
         _creation_counter=None,
-        **extra_args
+        **extra_args,
     ):
         super(InputField, self).__init__(_creation_counter=_creation_counter)
         self.name = name
         if required:
+            assert (
+                deprecation_reason is None
+            ), f"InputField {name} is required, cannot deprecate it."
             type_ = NonNull(type_)
         self._type = type_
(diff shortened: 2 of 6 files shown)


erikwrede:
That was fast! Thank you for the PR :) I'll check it out ASAP!

vhutov:
@erikwrede I felt some courage to play with `MountedType` and `SubclassWithMeta` support. Now it's generic and works with `Enum`'s, `InputObject`'s and `Scalar`'s. I added a class method which describes which Meta fields to inherit. I think it's clear to a reader what to expect. LMKWYT

vhutov:
@erikwrede Hey! Have you had a chance to take a look?

erikwrede:
@vhutov Sorry for taking to long, I was actually just looking at your new changes right now. 
Need to play around with it to give you a more thorough opinion though, as this is had become a larger change than just adding the deprecation now. Trying to get back to you on the weekend 🙂

vhutov:
@erikwrede no worries, I'm not in rush. I just thought it'd be better to add Meta support as well. I can certainly split them in two, if you think it's better. Thanks for looking at this!

erikwrede:
Having thought about it again, think it's better to split up the two PRs and think about the `mountable_meta` again. The current proposal would  also require users to adapt the new `mountable_meta` classmethod,  potentially breaking things by requiring them to implement `mountable_meta` for all `kwargs` they want to be inheritable.

If we are talking about a redesign of that part, I'd personally prefer to explicitly get the inheritable fields from a type, without having to implement a second `mountable_meta` method.
That configuration, it would be less redundant and field name changes wouldn't pose an issue if someone forgot to modify the return of `mountable_meta`.

A rough concept could look something like this:

Assume we have a base class (roughly based on the actual `InputTypeOptions` class in Graphene)
```python
class InputObjectTypeOptions:
   description: str
   deprecation_reason: str
   abcd: str
```
we would have an explicit definition of inheritable fields from that class. It is also extendable by subclassing it.

We could now adapt this for mounted types like `Argument`.
If a mounted Type specifies a class `MountedTypeOptions`, we can use that to extract all fields that are contained in the `MountedTypeOptions` from the `UnmountedTypeOptions` **iff** they exist on both types:

```python 
InputObjectType.meta_type = InputObjectTypeOptions
class ArgumentOptions:
   description: str
   deprecation_reason: str
   required: bool



Mount InputObjectType
# Extract deprecation reason from InputObjectTypeOptions (InputObjectType._meta)
# Don't extract required: field not defined on InputObjectTypeOptions
# Don't extract abcd: field not defined on ArgumentOptions
```
 If no `MountedTypeOptions` class exists for the mounted type,  we fall back to the legacy behavior.

This is just an initial idea, and by no means a finalized design. Also open to suggestions on why we should use `mountable_meta` instead. LMKWYT!

vhutov:
@erikwrede cleaned up the PR to contain only the deprecation reason ✅ 
Let me think about the proposed idea meanwhile.

vhutov:
**UPD:** Changed PR description

graphene/types/field.py
```diff
@@ -89,6 +89,9 @@ def __init__(
         ), f'The default value can not be a function but received "{base_type(default_value)}".'
 
         if required:
+            assert (
+                deprecation_reason is None
+            ), f"Field {name} is required, cannot deprecate it."
```
erikwrede, line 94: This restriction only applies to _input object fields_ & arguments : https://spec.graphql.org/draft/#sec--deprecated

graphene/types/tests/test_mutation.py
```diff
@@ -149,14 +148,14 @@ class MyMutation(ObjectType):
...
-            required=True,
+            required=False,
         )
 
     field = MyMutation._meta.fields["create_user"]
     assert field.name == "createUser"
     assert field.description == "Create a user"
     assert field.deprecation_reason == "Is deprecated"
-    assert field.type == NonNull(CreateUser)
+    assert field.type == CreateUser
```
erikwrede, line 159: See field.py

erikwrede (changes requested):
Review for this was still on my list - thank you for updating the PR!

I noticed that you made changes to `field.py` to only allow deprecation on nullable fields. According to the Spec, this constraint only applies to input fields. 
Look at the following example from spec:

```
type ExampleType {
  invalidField(
    newArg: String
    oldArg: String! @deprecated(reason: "Use `newArg`.")
  ): String
}
```
With input fields, the restriction makes sense, as we don't have any alternative to passing a value to the required `oldField`, although `oldField` is deprecated. For input fields, the alternative is to not query the field anymore.

The rest looks good, thank you!

erikwrede (approved):
LGTM!
As a follow-up, we should rework the `get_introspection_query()` for schema to allow for optional args to be introspected. Currently `schema.introspect()` calls this using the default args that set `input_value_deprecation: bool = False,`, excluding any deprecated fields or args.

erikwrede:
Will release tonight in `3.2.1`, thank you! 🙂

vhutov:
Thanks!

kashifalikhan36:
That's impressive

PR #1474, opened 2022-11-14
Disambiguate argument name in quickstart docs

### LINKED ISSUE #1230: [Documentation] Don't use "name" field for hello world example
**Is your feature request related to a problem? Please describe.**
I'm a new user and while reading through the getting started page of the docs, I was confused by the `name` input into the Scalar. `name` here acts as an input for the resolver, but `name` is also used to set a custom name for the field.

Here are the two different uses for `name` within the first 1 pages of docs:
https://docs.graphene-python.org/en/latest/quickstart/#creating-a-basic-schema
https://docs.graphene-python.org/en/latest/types/schema/#auto-camelcase-field-names

**Describe the solution you'd like**
Don't use the input `name` on the getting started page in the documentation. Use `fullName`, `firstName`, etc. instead.

**Describe alternatives you've considered**
Another option is to mention this behavior in the docs when introducing custom field names, but I feel like this would be less concise.

**Additional context**
image
image
Line 5 and line 3 of these images are what I'm referring to. These screenshots are from the links above.

rgroothuijsen:
As shown in #1230, the documentation has caused some confusion among new users because of `name` being simultaneously used as a keyword and as a field name. Given that most people will initially rely on the quickstart documentation, the `name` field name has been changed to something that's more obviously not a keyword.

Fixes #1230

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -37,12 +37,12 @@ An example in Graphene
 
 Let’s build a basic GraphQL schema to say "hello" and "goodbye" in Graphene.
 
-When we send a **Query** requesting only one **Field**, ``hello``, and specify a value for the ``name`` **Argument**...
+When we send a **Query** requesting only one **Field**, ``hello``, and specify a value for the ``firstName`` **Argument**...
 
 .. code::
 
     {
-      hello(name: "friend")
+      hello(firstName: "friend")
     }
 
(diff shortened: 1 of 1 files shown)


erikwrede (approved):

erikwrede:
Thank you! 🙂

PR #1475, opened 2022-11-14
Clarify execution order in middleware docs

### LINKED ISSUE #1385: Mention the order in which middlewares are executed in the docs
**Is your feature request related to a problem? Please describe.**

I was very surprised to find that graphene-python doesn't follow the same order for the execution of the middlewares, as [django](https://docs.djangoproject.com/en/3.2/topics/http/middleware/#middleware-order-and-layering):

> During the request phase, before calling the view, Django applies middleware in the order it’s defined in MIDDLEWARE, top-down.

For graphene, it goes bottom up.

**Describe the solution you'd like**
To document this behavior.

**Additional context**
I confirmed this order by printing to stdout from the `resolve` method of each middleware. It was opposite of the order in which I had specified them.

rgroothuijsen:
This PR adds a note to the middleware documentation noting the last-to-first execution order for middlewares, since #1385 indicates that some people were surprised by this behavior.

Fixes #1385

### DIFF
--- a/docs/execution/middleware.rst
+++ b/docs/execution/middleware.rst
@@ -41,6 +41,8 @@ And then execute it with:
 
     result = schema.execute('THE QUERY', middleware=[AuthorizationMiddleware()])
 
+If the ``middleware`` argument includes multiple middlewares,
+these middlewares will be executed bottom-up, i.e. from last to first.
 
 Functional example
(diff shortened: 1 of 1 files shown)


erikwrede (approved):
Thank you!

PR #1476, opened 2022-11-16
Remove unnecessary dependency on 'promise' library

### LINKED ISSUE #1296: Test Client is still promise based

mike-roberts-healx:
Previously, installing graphene and trying to do `from graphene.test import Client` as recommended in the docs caused an `ImportError`, as the 'promise' library is imported but only listed as a requirement in the 'test' section of the setup.py file.

Based on #1297 which appears to be inactive. Fixes #1296.

### DIFF
--- a/graphene/relay/tests/test_mutation_async.py
+++ b/graphene/relay/tests/test_mutation_async.py
@@ -3,6 +3,7 @@ from pytest import mark
 from graphene.types import ID, Field, ObjectType, Schema
 from graphene.types.scalars import String
 from graphene.relay.mutation import ClientIDMutation
+from graphene.test import Client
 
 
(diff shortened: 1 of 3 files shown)


erikwrede (approved):

erikwrede:
My enter key was stuck on merge, so here's a more complete merge Message for future reference:
`feat: Graphene TestClient supports async execution & Removes Promises which are no longer supported`

PR #1477, opened 2022-11-16
fix: MyPy findings due to a mypy version upgrade were corrected

erikwrede:
Fixes MyPy errors causing all pipelines to fail after an upgrade of the mypy version.


flipbit03 (approved):
LGTM!

PR #1486, opened 2022-12-16
Correct the module name of custom scalar example in documentation

pei-lun:
Update documentation because `graphql.language.ast.StringValue` in V2 has been renamed to `graphql.language.ast.StringValueNode` in V3.


erikwrede (approved):
Great Catch, thank you!

PR #1487, opened 2022-12-23
Remove prerelease notice

pederjohnsen:
I'm assuming v3 is no longer a prerelease and this notice should be removed 🤔  (the v2 link just takes you to the same docs so 🤷🏻‍♂️ )

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1,12 +1,6 @@
 Graphene
 ========
 
-------------
-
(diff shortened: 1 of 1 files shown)


erikwrede (approved):
Makes sense. We should implement a version switcher in the docs, but the website dependencies are really outdated so that's a larger refoctor. 
Thank you!

PR #1488, opened 2022-12-27
Add `subscribe` to Field

Luferov:
This PR makes possible to use external functions as subscriptions, not just `subsctibe_field`.

Before:

```python
class Query(ObjectType):
    hello = String()

    def resolve_hello(root, info):
        return "Hello, world!"

    count_to_ten = Field(Int)

    async def subscribe_count_to_ten(root, info): # <- you can only subscribe...
        for count in range(1, 11):
            yield count

schema = Schema(query=Query, subscription=Subscription)
```

After:

```python
class Query(ObjectType):
    hello = String()

    def resolve_hello(root, info):
        return "Hello, world!"


async def subscribe_count_to_five(root, info):
    for count in range(1, 6):
        yield count


    count_to_ten = Field(Int)

    async def subscribe_count_to_ten(root, info):
        for count in range(1, 11):
            yield count

    count_to_five = Field(Int, subscribe=subscribe_count_to_five) # <- Can use external function

schema = Schema(query=Query, subscription=Subscription)
```

At the same time, the permission function directly in the class has priority over Field.subscribe.

### DIFF
--- a/graphene/types/field.py
+++ b/graphene/types/field.py
@@ -51,6 +51,8 @@ class Field(MountedType):
             value object. If not set, the default resolver method for the schema is used.
         source (optional, str): attribute name to resolve for this field from the parent value
             object. Alternative to resolver (cannot set both source and resolver).
+        subscribe (optional, AsyncIterable): Asynchronous iterator to get the value for a Field from the parent
+            value object. If not set, the default resolver method for the schema is used.
         deprecation_reason (optional, str): Setting this value indicates that the field is
             depreciated and may provide instruction or reason on how for clients to proceed.
         required (optional, bool): indicates this field as not null in the graphql schema. Same behavior as
@@ -69,6 +71,7 @@ class Field(MountedType):
         args=None,
         resolver=None,
         source=None,
+        subscribe=None,
         deprecation_reason=None,
         name=None,
         description=None,
@@ -107,6 +110,7 @@ class Field(MountedType):
         if source:
             resolver = partial(source_resolver, source)
         self.resolver = resolver
+        self.subscribe = subscribe
         self.deprecation_reason = deprecation_reason
         self.description = description
         self.default_value = default_value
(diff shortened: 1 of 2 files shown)


erikwrede (commented):
Thanks for the PR! This could definitely come in handy, just like its equivalent for fields. 

Is there any specific reason to divert from the  procedure that incorporates the parent resolver we've used for classic field resolvers (see wrap resolve)?

The GQL Spec refers to the resolvers of subscriptions as resolvers: https://spec.graphql.org/draft/#sec-Source-Stream

I'd prefer should stick to that definition and use the resolver keyword argument for wrap subscribe as well. After all, a field cannot have a normal query resolver and a subscription resolver at the same time.

Luferov:
Thank you very much for your comment. This PR introduces the ability to use the feature to resolve subscriptions as Field resolve.
Initially, I thought so, but there were difficulties with the fact that subscribe and resolve are the same resolvers and it doesn't work:
https://github.com/graphql-python/graphene/blob/52143473efad141f6700237ecce79b22e8ff4e41/graphene/types/schema.py#L361-L368

Сan add a condition above:
```python
if subscribe is resolve:
    resolve = field_default_resolver
```
But I don't think this solution is beautiful.

The most adequate solution that I came up with is to separate asynchronous generators for subscriptions and other functions for ordinary resolvers.

graphene/types/field.py
```diff
@@ -121,18 +121,22 @@ def wrap_resolve(self, parent_resolver):
...
+        The Field resolver cannot be asynchronous generator.
         """
         if self.get_resolver is not None:
             warn_deprecation(
                 "The get_resolver method is being deprecated, please rename it to wrap_resolve."
             )
             return self.get_resolver(parent_resolver)
 
-        return self.resolver or parent_resolver
+        return not inspect.isasyncgenfunction(self.resolver) and self.resolver or parent_resolver
```
erikwrede, line 134: Why do we need to change this to include the inspection? A field can either be used for subscriptions or for normal queries (XOR). This also might break future support of `stream` or `defer` depending on how they'll be integrated

PR #1495, opened 2023-02-13
881: Corrected enum metaclass to fix pickle.dumps()

senseysensor:


senseysensor:
wow, seems fix doesn't work well for python less than 3.11. Will check with lower versions.

senseysensor:
update: this was not related to tox in anyway, but that was case with colliding enum class names defined in different modules (in different tests).
Proposed solution is not ideal but I think that would cover most cases of pickling.

erikwrede:
Thank you for your submission of the PR and for conducting a the investigation! I find it reasonable to enforce the uniqueness of Enum Names, as type names within a GQL schema are inherently unique. I would appreciate it if you could clarify whether this restriction solely applies to graphene Enums or if both Graphene Enums and normal Enums can have the same name.

I am also interested in exploring other methods of enhancing pickle support. Would a solution like the following be effective in addressing the issue?

```python
class EnumMeta:
	...
	def __reduce__(self):
	        return (self.__name__, tuple(self.__bases__), dict(self.__members__))
```

senseysensor:
Well, first of all, enforcing uniquness (of graphene enums only) is simple to implement, one more check can be added in `__new__` method to achieve this. In this case we don't need anything from my 2nd commit (https://github.com/graphql-python/graphene/pull/1495/commits/dac4ba4e1261526fb012955db7be0f5a4eb6b839). To be honest, this overcome from https://github.com/graphql-python/graphene/pull/1495/commits/dac4ba4e1261526fb012955db7be0f5a4eb6b839 is not reliable (will work only for some specific cases)  and includes black magic from inspect module which in most cases smells. I probably will revert back to https://github.com/graphql-python/graphene/pull/1495/commits/f5b5a14f3f35eb215b9592e456cb5018044ea39a but will assign unique name of the class (across other tests).

Also not sure should we put hard restriction in creation non-unique graphene Enums or not, because the only side effect from this is that pickling wouldn't work, which is not so wide-used feature, and all other functionality works just fine. On the other hand, if we don't, user potentially may receive weird errors while pickling.

Limitation of unique Enum classes goes only for graphene enums and doesn't affect native python Enums in any way.

About your last question regarding better enum support and use some other methods like `__reduce__` or `__reduce_ex__` – this won't work, unfortunately. That was my first thought here, but source of the issue is not that graphen enum itself is not pickable, it perfectly is (it's just wrapper around native Enums, which supports pickle). But cause is the way how graphene enums are created, enum's direct metaclass is dynamically created / modified and thus pickle fails found its type in graphene module during its inner preliminary checks.

So to sum this up, I will commit another changes which revert code back to commit 1, but will rename class to get it unique to make tests pass.

senseysensor:
@erikwrede , should be something else get changed here or that could be merged..?

erikwrede:
Hey, thanks for the patience. 
I'm unsure about modifying `globals` in the `Metaclass` and need to do some research on it first. Could this cause naming conflicts? (e.g. when you have a non-graphene Enum `MyEnum` and then generate a graphene Enum also named `MyEnum` from it)

senseysensor:
Well, I probably am not completely sure about your considerations about possible conflicts. At least in specific example you gave that shouldn't be any conflict. `globals()` refers not to metaclass but to this specific module (`__dict__` of
 `graphene/types/enum.py`).
Native Enum class has its metaclass defined in standard library `enum.EnumType`. It doesn't care about graphene Enum in any way (graphene encapsulate native one). The only trick is done here is adding reference of dynamically created metaclass ( hypotethical `GrapheneEnum`) to `graphene/types/enum.py` module so that pickle dump algorithm is not get confused during internal checks.
So the only conflict I can consider is when we create several graphene Enums with same name in scope of runtime. In that case 1st item we added into `__dict__` of the module will get overridden with 2nd definition, and pickle would throw another error during dump (anyway that should affect only pickle behavior).

erikwrede (approved):

Issue #1454, opened 2022-08-27
Correct documentation and use of type hints


erikwrede:
We should upgrade graphene to use modern type hints. In that process, type hints should be removed from the class/method documentation to have the code as the single source of truth for typing. 
We can use https://github.com/tox-dev/sphinx-autodoc-typehints to automatically add type hints back into the docs.

Known issues: 
#1100 , see #1282

anentropic:
This would big a big help to anyone wanting to use graphene in a typed codebase

Because the 3rd party type defs we've been using seem to be abandonware (e.g. https://github.com/trialspark/graphene-stubs/pull/27 and no commits for ~3 years)

PR #1499, opened 2023-03-02
chore: make relay type fields extendable

erikwrede:
This PR enables users to easily modify the `relay.Connection`  type to add fields like `nodes`, or customize the existing `Edges` and  `PageInfo` fields to make changes about nullability. 
This is handy for cases like having a generic nodes field, that skips the use of edges for non-pagination use. The change is enabled by checking if `_meta` was already provided in the keyword arguments of `Node/Connection`'s `init_subclass_with_meta`, so we don't override existing field definitions from the superclass.

The most common use case will be to add a generic `nodes` field for cases where cursor-based pagination is not necessary:

```graphql
type MyTypeConnection {
	pageInfo: PageInfo
	edges: [MyTypeEdge]
	nodes: [MyType]
}

### DIFF
--- a/graphene/relay/connection.py
+++ b/graphene/relay/connection.py
@@ -26,11 +26,11 @@ def get_edge_class(
 
     edge_name = f"{base_name}Edge"
     if edge_class:
-        edge_bases = (edge_class, EdgeBase, ObjectType)
+        edge_bases = [edge_class, EdgeBase, ObjectType]
     else:
-        edge_bases = (EdgeBase, ObjectType)
+        edge_bases = [EdgeBase, ObjectType]
 
-    return type(edge_name, edge_bases, {"Meta": EdgeMeta})
+    return type(edge_name, tuple(edge_bases), {"Meta": EdgeMeta})
 
 
 class PageInfo(ObjectType):
--- a/graphene/relay/connection.py
+++ b/graphene/relay/connection.py
@@ -26,7 +26,7 @@ def get_edge_class(
 
     edge_name = f"{base_name}Edge"
 
-    edge_bases = [edge_class] if edge_class else [EdgeBase]
+    edge_bases = [edge_class, EdgeBase] if edge_class else [EdgeBase]
     if not isinstance(edge_class, ObjectType):
         edge_bases = [*edge_bases, ObjectType]
 
--- a/graphene/relay/connection.py
+++ b/graphene/relay/connection.py
@@ -61,8 +61,9 @@ class Connection(ObjectType):
         abstract = True
 
     @classmethod
-    def __init_subclass_with_meta__(cls, node=None, name=None, **options):
-        _meta = ConnectionOptions(cls)
+    def __init_subclass_with_meta__(cls, node=None, name=None, _meta=None, **options):
+        if not _meta:
+            _meta = ConnectionOptions(cls)
         assert node, f"You have to provide a node in {cls.__name__}.Meta"
         assert isinstance(node, NonNull) or issubclass(
             node, (Scalar, Enum, ObjectType, Interface, Union, NonNull)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by MattKleinsmith** [comment]
Thank you. This is cool.

I don't feel qualified to approve or deny, but I do like the design.

File: graphene/relay/connection.py [hunk 1]
```diff
@@ -1,14 +1,36 @@
...
+        cursor = String(required=True, description="A cursor for use in pagination")
+
+    class EdgeMeta:
+        description = f"A Relay edge containing a `{base_name}` and its cursor."
+
+    edge_name = f"{base_name}Edge"
+    if edge_class:
+        edge_bases = [edge_class, EdgeBase, ObjectType]
+    else:
+        edge_bases = [EdgeBase, ObjectType]
```
jendrikjoe, line 31: Could you elaborate why you explicitly add `[edge_class, EdgeBase, ObjectType]` or `[EdgeBase, ObjectType]`, respectively as edge_bases? 🙂 Wouldn't it be easier to let `EdgeBase` inherit from `ObjectType` and check that edge_class does the same? 🙂 Or alternatively put `EdgeBase` into global scope and ensure that if `edge_class` is set it inherits from `EdgeBase`? 🙂 Would definitely be a stronger restriction on the user side, but I am not sure if it can yield unintended side effects when adding `EdgeBase` and `ObjectType` as additional bases when the user assumes only `edge_class` is being used 🙂

jendrikjoe (commented):
So helpful going forward 👍 This will scrap a lot of workarounds from a lot of peoples codebases 🥳

graphene/relay/connection.py
```diff
@@ -61,8 +83,9 @@ class Meta:
         abstract = True
 
     @classmethod
-    def __init_subclass_with_meta__(cls, node=None, name=None, **options):
-        _meta = ConnectionOptions(cls)
+    def __init_subclass_with_meta__(cls, node=None, name=None, _meta=None, **options):
```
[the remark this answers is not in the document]
erikwrede: I believe there's not much we can change about that due to the Architecture of `SubclassWithMeta`. Luckily the `graphene` mypy check isn't complaining. This is a related issue for typing: #1454

graphene/relay/connection.py
erikwrede: I honestly just kept the old behavior to not induce any breaking changes. However, I agree that putting `EdgeBase` into Global Scope helps here! 
If we require inheritance from `EdgeBase`, or expect that in any way, we would create a breaking change :/

graphene/relay/connection.py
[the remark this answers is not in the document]
jendrikjoe: This is indeed covered in #1454 👍

graphene/relay/connection.py
jendrikjoe: I am generally fine with moving this to a separate issue. Alternatively we could do an 
```
edge_bases = [edge_class] if edge_class else [EdgeBase]
if not isinstance(edge_class, ObjectType):
    edge_bases = [*edge_bases, ObjectType]
```
This way we don't assume inheritance any inheritance and would add `ObjectType` if it is not inherited from. Alternatively, we could do this with `EdgeBase` if we actually require it to be in edge_bases, but I think ObjectType should be enough right? 🙂

erikwrede: Done 🙂

graphene/relay/connection.py
```diff
@@ -25,10 +25,10 @@ class EdgeMeta:
         description = f"A Relay edge containing a `{base_name}` and its cursor."
 
     edge_name = f"{base_name}Edge"
-    if edge_class:
-        edge_bases = [edge_class, EdgeBase, ObjectType]
-    else:
-        edge_bases = [EdgeBase, ObjectType]
+
+    edge_bases = [edge_class] if edge_class else [EdgeBase]
```
jendrikjoe: My bad in not being specific with my comment 🙂  If EdgeBase doesn't inherit from `ObjectType`. This should of course be `edge_bases = [edge_class] if edge_class else [EdgeBase, ObjectType]` 🙂

graphene/relay/connection.py
```diff
@@ -26,7 +26,7 @@ class EdgeMeta:
 
     edge_name = f"{base_name}Edge"
 
-    edge_bases = [edge_class] if edge_class else [EdgeBase]
+    edge_bases = [edge_class, EdgeBase] if edge_class else [EdgeBase]
```
jendrikjoe, line 29: Do we want to add `EdgeBase` to edge_class as a base, or do we just want to put it upon the user to have node and cursor in their definition of edge class? 🙂 Simply not sure if this would be the behavior a user expects 🙂

graphene/relay/connection.py
jendrikjoe: 👍

graphene/relay/connection.py
erikwrede: Ideally we don't want to add this here, but we have to do so  in order not to create a breaking change.

The workaround is to create a custom `Connection` that does not use this function and instead doesn't put  `EdgeBase`  into the custom `Edge` bases - hacky, but it works

jendrikjoe: Ahh okay got it 👍 The let's keep it as is 👍

graphene/relay/connection.py
erikwrede: Should be resolved now, right? The next line practically does that for both.

jendrikjoe: Ohh sure 🙈

jendrikjoe (approved):

PR #1502, opened 2023-04-11
Default enum description to "An enumeration."

firaskafri:
Default enum description to "An enumeration." string as opposed to trying to read `enum.__doc__` as it appears to be not set to default enum class doc in python 3.11

### DIFF
--- a/graphene/types/enum.py
+++ b/graphene/types/enum.py
@@ -63,7 +63,7 @@ class EnumMeta(SubclassWithMeta_Meta):
         cls, enum, name=None, description=None, deprecation_reason=None
     ):  # noqa: N805
         name = name or enum.__name__
-        description = description or enum.__doc__
+        description = description or "An enumeration."
         meta_dict = {
             "enum": enum,
             "description": description,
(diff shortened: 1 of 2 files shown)


graphene/types/enum.py
```diff
@@ -63,7 +63,7 @@ def from_enum(
         cls, enum, name=None, description=None, deprecation_reason=None
     ):  # noqa: N805
         name = name or enum.__name__
-        description = description or enum.__doc__
+        description = description or "An enumeration."
```
sjdemartini, line 66: How about:

```python
description = description or enum.__doc__ or "An enumeration."
``` 
instead, to still be able to utilize the docstring descriptions of enums when they _are_ present, like:

```python
>>> class Color(Enum):
...     """my description"""
...     RED = 1
...
>>> Color.__doc__
'my description'
```

firaskafri: Awesome!

erikwrede:
Thanks for the PR @firaskafri 🙂 
Can we add one or two tests for this that check the prioritization (Description over enum.__doc__ over "An Enumeration")? 🙂

firaskafri:
> Thanks for the PR @firaskafri 🙂 Can we add one or two tests for this that check the prioritization (Description over enum.**doc** over "An Enumeration")? 🙂

Hi @erikwrede !

Just added two tests to address these cases!

erikwrede (dismissed):

erikwrede (approved):

PR #1503, opened 2023-04-14
fix: Use `typing.TYPE_CHECKING` to provide generalized way

rapsealk:
Hello graphene!

I found that the code base is using `MYPY` as a constant to tell whether a type checker is running.

According to the [mypy docs](https://mypy.readthedocs.io/en/stable/runtime_troubles.html#typing-type-checking), however, `mypy` introduces a more commonly used constant: `typing.TYPE_CHECKING`. So I'd like to suggest using this rather than `MYPY`.

As `typing.TYPE_CHECKING` has been added since Python 3.5.2 and `graphene` supports from Python 3.6, there would be no problem with backporting.

Reference:
- https://docs.python.org/ko/3/library/typing.html#typing.TYPE_CHECKING

### DIFF
--- a/graphene/types/inputobjecttype.py
+++ b/graphene/types/inputobjecttype.py
@@ -1,11 +1,12 @@
+from typing import TYPE_CHECKING
+
 from .base import BaseOptions, BaseType
 from .inputfield import InputField
 from .unmountedtype import UnmountedType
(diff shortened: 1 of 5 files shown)


erikwrede:
Thanks! 😊

erikwrede (approved):

PR #1504, opened 2023-05-04
types: add option for strict connection types

shrouxm:
This PR adds a `strict_types` meta option to the relay `Connection` class to simplify the use case identified in this issue: https://github.com/graphql-python/graphene-django/issues/901

Default of `False` is the old behavior, and when set to `True` generated `TConnection` types will be typed as `edges: [TEdge!]!` instead of `edges: [TEdge]!` and `TEdge` will have `node: T!` instead of `node: T`.

This avoids duplicating logic and string descriptions for subclasses which just want to implement this common behavior. If this behavior is deemed desirable by default it would be trivial to flip the flag's default or remove it entirely. It only make the types stricter so it oughtn't break any existing code (though it will cause unexpected changes to user's schema's nonetheless, which is why I kept the default to `False`).

### DIFF
--- a/graphene/relay/connection.py
+++ b/graphene/relay/connection.py
@@ -22,8 +22,8 @@ def get_edge_class(
 
     class EdgeBase:
         node = Field(
-            NonNull(_node) if strict_types else _node, 
-            description="The item at the end of the edge"
+            NonNull(_node) if strict_types else _node,
+            description="The item at the end of the edge",
         )
         cursor = String(required=True, description="A cursor for use in pagination")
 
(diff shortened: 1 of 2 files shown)


erikwrede (changes requested):
Thanks for the PR! Can you please add some test cases for this, testing the schema generation? 😊

shrouxm:
@erikwrede does that look like what you meant?

shrouxm:
@erikwrede it looks like the checks will never complete because the python 3.6 test workflow no longer exists after https://github.com/graphql-python/graphene/pull/1507/files but the GitHub repo settings still require that workflow to run

erikwrede (approved):

PR #1509, opened 2023-06-22
Apply PEP 681 dataclass_transform to graphene ObjectTypeMeta

roshanjrajan-zip:
Following Eric Traut's advice here - https://github.com/microsoft/pyright/issues/5277#issuecomment-1588089626

This is a typing only fix that doesn't change any code behavior but helps type checkers. Essentially for pyright version 1.1.307+, when creating graphene.ObjectType that is returned, pyright will fail saying there are `Expected no arguments to "SomeObject" constructor`. This fixes the error and allows these members to be typed.

### DIFF
--- a/graphene/types/objecttype.py
+++ b/graphene/types/objecttype.py
@@ -1,4 +1,5 @@
-from typing import TYPE_CHECKING, dataclass_transform
+from typing import TYPE_CHECKING
+from typing_extensions import dataclass_transform
 
 from .base import BaseOptions, BaseType, BaseTypeMeta
 from .field import Field
--- a/graphene/types/objecttype.py
+++ b/graphene/types/objecttype.py
@@ -1,4 +1,4 @@
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, dataclass_transform
 
 from .base import BaseOptions, BaseType, BaseTypeMeta
 from .field import Field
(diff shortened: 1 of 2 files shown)


roshanjrajan-zip:
@erikwrede Could you take a look at this PR?

erikwrede:
@roshanjrajan-zip its on my list, thanks for noticing and working on a fix 🙂

roshanjrajan-zip:
Okay I didn't realized but `dataclass_transform` is supported for 3.11+. I can push a change for 3.11 support and then flag it behind a 3.11 check?

erikwrede:
@roshanjrajan-zip we can import it from `typing_extensions`

roshanjrajan-zip:
@erikwrede Updated the PR!

roshanjrajan-zip:
@erikwrede Did you have a chance to look at this PR?

roshanjrajan-zip:
Hi @erikwrede! Sorry to ping you so much. Wondering if there was any other feedback on this PR?

erikwrede:
@roshanjrajan-zip I'm tending towards approving this, but I'm unsure if this will solve our problems. The arguments to the constructor will continue to be recognized, however; our code won't work the way it is suggested in the issue:

> ```python
> a = TestObject(var1=Int(0), var2=String("hello"))
> print(a.var1, a.var2)
> ```

Graphene Objects are always initialized like this:

```python
a = TestObject(var1=0, var2="hello")
```

So there will still be a type conflict on each constructor argument. I need to discuss this with the other maintainers first, which can take some time, hence the long wait. LMK what you think.

roshanjrajan-zip:
@erikwrede I was thinking about the same problem because it requires a bunch of manual fixes from the my side to support this behavior. I was going to codemod my code to make this change but open to other options!

RJPercival:
I think this was a step in the right direction. It just needs to be followed by a mypy plugin similar to [django-stubs](https://github.com/typeddjango/django-stubs) that converts the field types to their corresponding underlying types (e.g. `String` to `str`). Of course, that plugin would only work for mypy, but that's going to cover a lot of people's needs.

PR #1511, opened 2023-07-04
Simplify props.py

RJPercival:
It had code for handling the differences between old-style and new-style classes in Python 2, but support for Python 2 was dropped long ago.

This could be simplified further, e.g. by skipping any attributes that start with an underscore, but that would be a behavioural change, whereas this PR is a pure refactoring.

### DIFF
--- a/graphene/utils/props.py
+++ b/graphene/utils/props.py
@@ -1,15 +1,11 @@
-class _OldClass:
+class _Class:
     pass
 
 
-class _NewClass:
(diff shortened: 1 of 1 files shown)


erikwrede:
Thanks for the PR @RJPercival 🙂 
Could you please run the pre-commit hooks before commiting, so that the linter runs?  Feel free to tag me after the PR is updated, will merge then. 😊

PR #1514, opened 2023-07-28
Add Email type

arunpkm:
Add Email type

### DIFF
--- a/graphene/types/email.py
+++ b/graphene/types/email.py
@@ -1,0 +1,36 @@
+from __future__ import absolute_import
+
+
+from graphql import Undefined
+from graphql.language.ast import StringValueNode
+import json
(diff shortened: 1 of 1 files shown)


erikwrede:
Hey @arunpkm, what exactly are you trying to achieve with this PR? While the addition of a custom Email Scalar might make sense for a clean schema design, I'm not sure wether this should be part of the graphene implementation.
The specifics of this scalar may vary from company to company as there may be more granular email rules than those defined in the RFCs mentioned by you. Consequently, it might make sense to leave this to the individual developers.

Aside from that, it is important to provide test cases and up to date documentation in the PR. Currently, the scalar doesn't seem to be doing any email validation. What are your plans for this, going forward? LMK what you think 😊

PR #1515, opened 2023-08-08
fix: snake case converter corner cases

tcleonard:
There were some inconsistencies in the way the snake case conversion was working.

1. the handling of string already containing `_` was not consistent. Indeed in the old implementation:
```python
assert to_snake_case("_File__") == "__file__"
assert to_snake_case("_IPhoneHysteria") == "_i_phone_hysteria"  # new implementation "__i_phone_hysteria"
assert to_snake_case("_A0") == "_a0"  # new implementation "__a0"
assert to_snake_case("toto_Ao") == "toto__ao"
assert to_snake_case("toto_A0") == "toto_a0"  # new implementation "toto__a0"
```
2. the handling of consecutive capitals was not consistent. Indeed in the old implementation:
```python
assert to_snake_case("SnakesOnAPlane") == "snakes_on_a_plane"
assert to_snake_case("SNakesOnAPlane") == "s_nakes_on_a_plane"
assert to_snake_case("SN0w") == "sn0w"  # new implementation "s_n0w"
assert to_snake_case("SNow") == "s_now"
assert to_snake_case("AAA") == "aaa"  # new implementation "a_a_a"
assert to_snake_case("AAAa") == "aa_aa"  # new implementation "a_a_aa"
```

### DIFF
--- a/graphene/utils/str_converters.py
+++ b/graphene/utils/str_converters.py
@@ -10,8 +10,8 @@ def to_camel_case(snake_str):
     return components[0] + "".join(x.capitalize() if x else "_" for x in components[1:])
 
 
-# From this response in Stackoverflow
-# http://stackoverflow.com/a/1176023/1072990
 def to_snake_case(name):
-    s1 = re.sub("(.)([A-Z][a-z]+)", r"\1_\2", name)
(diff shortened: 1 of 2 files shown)


erikwrede:
This is a breaking change, how should we handle this? I believe there should be a feature toggle which we gracefully switch over.

PR #1520, opened 2023-08-30
remove pytz

wongcht:
`pytz` is being deprecated, the usage of pytz in the package can be easily replaced by `datetime.timezone.utc`

I also plan to apply this commit to v2.


erikwrede (approved):

erikwrede:
Thanks for all your PRs @wongcht! I will check the rest out in the weekend 😊

PR #1521, opened 2023-08-30
Support Python 3.8-3.11 only, with default 3.11

wongcht:
python 3.7 is already EOL, suggest to keep supported version as official 3.8-3.11, and set 3.11 as default in the pipeline


erikwrede:
Thanks for your PR, this was fixed recently in another PR

PR #1530, opened 2023-11-22
Get rid of Deprecation Warnings

patillacode:
In my current project which depends on `graphene` (3.3) I encountered my logs unusable since they were heavily polluted with the following warning:

```
/home/wagtail/.local/lib/python3.9/site-packages/graphene/utils/subclass_with_meta.py:46: DeprecationWarning: Creating a DjangoObjectType without either the `fields` or the `exclude` option is deprecated. Add an explicit `fields = '__all__'` option on DjangoObjectType CollectionObjectType to use all fields
```

While trying to fix it just by adding the `fields="__all__"` option I stumbled upon this other one:

```
/home/wagtail/.local/lib/python3.9/site-packages/graphene/utils/subclass_with_meta.py:46: DeprecationWarning: Defining `exclude_fields` is deprecated in favour of `exclude`.
```

I believe this PR would make those warnings disappear without breaking anything AFAIK, at least that is what I found while testing it.

### DIFF
--- a/graphene/utils/subclass_with_meta.py
+++ b/graphene/utils/subclass_with_meta.py
@@ -40,7 +40,13 @@ class SubclassWithMeta(metaclass=SubclassWithMeta_Meta):
                 "Abstract types can only contain the abstract attribute. "
                 f"Received: abstract, {', '.join(options)}"
             )
+
         else:
+            if options.get("exclude_fields", None):
+                options["exclude"] = options.pop("exclude_fields")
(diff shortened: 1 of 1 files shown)


ssyberg:
Would it be possible to get this merged?

erikwrede:
Sorry for the late response. To me it looks like this is an issue that should be fixed in graphene_django, namely removing the use of a deprecated field in graphene. I'd prefer a solution that uses the newest standard in graphene instead of un-deprecating the deprecated option in the core library. Can you open a PR there?

Issue #1539, opened 2024-03-17
Tests in `graphene/relay/tests/test_custom_global_id.py` fail due incompatibility with pytest 8.x


tijuca:
* **What is the current behavior?**

Running the test suite, in particular the tests in `graphene/relay/tests/test_custom_global_id.py` fail if pytest 8.x is used.

```shell
__________ ERROR at setup of TestUUIDGlobalID.test_str_schema_correct __________

cls = <class '_pytest.runner.CallInfo'>
func = <function call_runtest_hook.<locals>.<lambda> at 0x7fdc59d0b920>
when = 'setup'
reraise = (<class '_pytest.outcomes.Exit'>, <class 'KeyboardInterrupt'>)

    @classmethod
    def from_call(
        cls,
        func: Callable[[], TResult],
        when: Literal["collect", "setup", "call", "teardown"],
        reraise: Optional[
            Union[Type[BaseException], Tuple[Type[BaseException], ...]]
        ] = None,
    ) -> "CallInfo[TResult]":
        """Call func, wrapping the result in a CallInfo.
    
        :param func:
            The function to call. Called without arguments.
        :param when:
            The phase in which the function is called.
        :param reraise:
            Exception or exceptions that shall propagate if raised by the
            function, instead of being wrapped in the CallInfo.
        """
        excinfo = None
        start = timing.time()
        precise_start = timing.perf_counter()
        try:
>           result: Optional[TResult] = func()

/usr/lib/python3/dist-packages/_pytest/runner.py:342: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/lib/python3/dist-packages/_pytest/runner.py:263: in <lambda>
    lambda: ihook(item=item, **kwds), when=when, reraise=reraise
/usr/lib/python3/dist-packages/pluggy/_hooks.py:501: in __call__
    return self._hookexec(self.name, self._hookimpls.copy(), kwargs, firstresult)
/usr/lib/python3/dist-packages/pluggy/_manager.py:119: in _hookexec
    return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
/usr/lib/python3/dist-packages/_pytest/unraisableexception.py:85: in pytest_runtest_setup
    yield from unraisable_exception_runtest_hook()
[...]

PR #1540, opened 2024-03-24
pytest: Don't use nose like syntax in graphene/relay/tests/test_custom_global_id.py (#1539)

tijuca:
The tests in test_custom_global_id.py use the old nose specific method 'setup(self)' which isn't supported anymore in Pytest 8+.

https://daobook.github.io/pytest/deprecations.html#setup-teardown


erikwrede:
thanks for the pr!

erikwrede (approved):

PR #1548, opened 2024-05-15
remove aniso8601, mock, iso8601

dulmandakh:
Python 3.7 added **fromisoformat** functionality, so we can remove aniso8601. There are test cases for it, and all are passing..

Removing **mock**, because the documentation states below, so we don't need it.
> mock is now part of the Python standard library, available as [unittest.mock (https://docs.python.org/dev/library/unittest.mock.html) in Python 3.3 onwards.


dulmandakh:
@syrusakbary please review and merge.

PR #1559, opened 2024-06-24
remove README.rst, leave only README.md

dulmandakh:
This PR removes README.rst as it's outdated version of README.md, also creates friction between GitHub and PyPI.


dulmandakh:
@erikwrede please review and merge. Let's remove outdated, perhaps unmaintained README.

PR #1565, opened 2024-06-29
migrate to pyproject.toml

dulmandakh:
This PR modernizes Python packaging usign pyproject.toml format.

Depends on:
- https://github.com/graphql-python/graphene/pull/1564 
- https://github.com/graphql-python/graphene/pull/1563


erikwrede:
I'd personally prefer switching to poetry or pdm right away when we make the switch to pyproject.toml What do you think? @dulmandakh

dulmandakh:
I use poetry for my projects. I can help with migration too.

I'll change the PR to use poetry then 😉

dulmandakh:
@erikwrede please review 😉

pyproject.toml
```diff
@@ -0,0 +1,43 @@
...
+
+[tool.poetry.group.test.dependencies]
+tox = "*"
+pytest = ">=8,<9"
+pytest-benchmark = ">=4,<5"
+pytest-cov = ">=5,<6"
+pytest-mock = ">=3,<4"
+pytest-asyncio = ">=0.16,<2"
+snapshottest = ">=0.6,<1"
+coveralls = ">=3.3,<5"
```
ogenstad, line 40: I think you missed ruff here which was added in another part of the setup.py file. Not sure if a separate group for "dev" is required in the same way as it existed in setup.py.

dulmandakh: I'm moving Ruff to pre-commit hook for lint consistency.

ogenstad: Ah ok, though that would depend on developers having pre commit hooks enabled and you wouldn't validate if this has been done within the CI.

I think it can be nice to be able to install everything that is needed within the project so that you can configure the editor to run formatters on save etc.

But this was only a minor comment if there was some oversight.

dulmandakh: Thank you for suggestion. I have an idea to refactor how lint is run, perhaps not rely on pre-commit as its purpose is just not check but fix issues during commit.

graphene/__init__.py
```diff
@@ -46,7 +46,6 @@
 from .utils.resolve_only_args import resolve_only_args
 
 __all__ = [
-    "__version__",
```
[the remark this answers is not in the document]
dulmandakh: I'll look into it and make changes accordingly

PR #1570, opened 2024-07-06
bump mypy to 1.10

dulmandakh:
This PR bumps mypy, and fixes below error.

```python
graphene/utils/deprecated.py:30: error: If x = b'abc' then f"{x}" or "{}".format(x) produces "b'abc'", not "abc". If this is desired behavior, use f"{x!r}" or "{!r}".format(x). Otherwise, decode the bytes  [str-bytes-safe]
graphene/utils/deprecated.py:32: error: If x = b'abc' then f"{x}" or "{}".format(x) produces "b'abc'", not "abc". If this is desired behavior, use f"{x!r}" or "{!r}".format(x). Otherwise, decode the bytes  [str-bytes-safe]
```

### DIFF
--- a/graphene/utils/deprecated.py
+++ b/graphene/utils/deprecated.py
@@ -27,9 +27,9 @@ def deprecated(reason):
 
         def decorator(func1):
             if inspect.isclass(func1):
-                fmt1 = f"Call to deprecated class {func1.__name__} ({reason!r})."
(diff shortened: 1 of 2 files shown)


dulmandakh:
@erikwrede please review and merge 🙏

erikwrede:
thank you, incorporated and superseded by #1578

PR #1572, opened 2024-07-16
Actually run the tests in python 3.12 and 3.13 and remove `snapshottest` dependency

PoByBolek:
I noticed that 614449e65193252209378cdf5bcbb97e20ab08be (PR #1550) introduced an invalid tox environment: `py12` instead of `py312`. This mistake was later repeated in 68343857868b08d4ff55153fc93fb6398196f8dc (PR #1561): `py13` instead of `py313`.

On my machine, `tox list` outputs:
```
default environments:
py38       -> [no description]
py39       -> [no description]
py310      -> [no description]
py311      -> [no description]
py312      -> [no description]
py313      -> [no description]
mypy       -> [no description]
pre-commit -> [no description]

additional environments:
py12       -> [no description]
py13       -> [no description]
```

And `tox run` doesn't run any tests in the `py312` and `py313` environments. This is also true in the Github workflows.

For example, in [this "successful" run in the `py312` environment from 2024-07-01](https://github.com/graphql-python/graphene/actions/runs/9746613624/job/26897390884) tox only runs `pip install .[test]` and nothing else.
grafik

This looks very different from [this run in the `py311` environment](https://github.com/graphql-python/graphene/actions/runs/9746613624/job/26897391277), which also runs `pytest`:
grafik

-----

This PR actually runs the tests in the `py312` and `py313` and drops the `snapshottest` dependency (which was only used in the example tests) so that the tests also work in Python 3.12 and above. This is necessary because [`snapshottest`](https://github.com/syrusakbary/snapshottest) (which appears [dead](https://github.com/syrusakbary/snapshottest/issues/174)) is using Python's `imp` module, [which has been deprected since 3.4 and was removed in 3.12](https://docs.python.org/3.11/library/imp.html).


PoByBolek:
From what I can tell there are at least two open pull requests in [`snapshottest`](https://github.com/syrusakbary/snapshottest) that could fix our test failures in `py312` and `py313`:
- https://github.com/syrusakbary/snapshottest/pull/173
- https://github.com/syrusakbary/snapshottest/pull/168

PoByBolek:
I also dropped the `snapshottest` dependency and simply moved the expectactions from the `snap_*.py` files to the corresponding test cases in the example tests. So the tests should now work across all environments again 😄

dulmandakh:
I did notice that tests on Python 3.12 are not working as expected, but hadn't chance to dig into it. Thank you 🙏

I agree that Graphene needs to drop snapshottest as it's failing to work on Python 3.12 and above. I'll try to take time and review your changes.

dulmandakh:
LGTM

dulmandakh (approved):
LGTM

PoByBolek:
@erikwrede ping?

erikwrede (approved):
Sorry for the delay, currently really busy in my day-to-day. 

Thanks for picking this up and pinging me!

erikwrede:
Seems like we have some test failures in 3.13. Will investigate later.

PoByBolek:
looks like the `json` module raises a new error message for trailing commas since https://github.com/python/cpython/commit/cfa25fe3e39e09612a6ba8409c46cf35a091de3c (Python 3.13.0a3)

graphene/types/tests/test_json.py
```diff
@@ -51,35 +51,30 @@ def test_jsonstring_invalid_query():
...
 
     result = schema.execute("""{ json(input: "{\\"key\\": 0,}") }""")
-    assert result.errors
     assert len(result.errors) == 1
-    assert result.errors[0].message == (
-        'Expected value of type \'JSONString\', found "{\\"key\\": 0,}"; '
-        "Badly formed JSONString: Expecting property name enclosed in double quotes: line 1 column 11 (char 10)"
+    assert result.errors[0].message.startswith(
+        'Expected value of type \'JSONString\', found "{\\"key\\": 0,}"; Badly formed JSONString:'
     )
```
PoByBolek, line 80: We could also specify the expected error message based on the python version. Something like
```python
if sys.version_info >= (3, 13):
    reason = "Illegal trailing comma before end of object: line 1 column 10 (char 9)"
else:
    reason = "Expecting property name enclosed in double quotes: line 1 column 11 (char 10)"
assert result.errors == [
    {"message": f'Expected value of type \'JSONString\', found "{{\\"key\\": 0,}}"; Badly formed JSONString: {reason}'},
]
```

But using `startswith()` here feels more robust to me.

erikwrede: Fine by me, we're testing for JSON Scalar handling, not for JSON Parsing here

erikwrede (approved):

erikwrede:
Thank you!

PR #1582, opened 2024-11-07
Do not raise AttributeError when parsing non-string UUIDs

phihag:
When a user sends a dictionary or other object as a UUID variable like `{[123]}`, previously graphene crashed with an `AttributeError`, like this:

```
(…)
  File "…/lib/python3.12/site-packages/graphql/utils/is_valid_value.py", line 78, in is_valid_value
    parse_result = type.parse_value(value)
                   ^^^^^^^^^^^^^^^^^^^^^^^
  File "…/lib/python3.12/site-packages/graphene/types/uuid.py", line 33, in parse_value
    return _UUID(value)
           ^^^^^^^^^^^^
  File "/usr/lib/python3.12/uuid.py", line 175, in __init__
    hex = hex.replace('urn:', '').replace('uuid:', '')
          ^^^^^^^^^^^
AttributeError: 'dict' object has no attribute 'replace'
```

But an `AttributeError` makes it seem like this is the server's fault, when it's obviously the client's.

Report a proper `GraphQLError`.

### DIFF
--- a/graphene/types/tests/test_uuid.py
+++ b/graphene/types/tests/test_uuid.py
@@ -47,7 +47,7 @@ def test_uuidstring_invalid_argument():
     assert len(result.errors) == 1
     assert (
         result.errors[0].message
-        == "Variable '$uuid' got invalid value {'not': 'a string'}; UUID must be a string"
+        == "Variable '$uuid' got invalid value {'not': 'a string'}; UUID cannot represent value: {'not': 'a string'}"
     )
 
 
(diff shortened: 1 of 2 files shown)


erikwrede:
Good catch! I adjusted your PR to keep allowing tuple-representations of UUIDs as well, and kept the error message consistent. Thanks for the Contribution 😊

erikwrede (approved):

erikwrede (approved):

PR #1594, opened 2025-02-10
[FIX] Decimal scalar inaccurate conversion of float to decimal

### LINKED ISSUE #1593: Inaccurate Float-to-Decimal Conversion in `parse_value` of `Decimal` `Scalar`
## Current Behavior
The `parse_value` function currently converts the input using `_Decimal(value)`. When a float is passed as input, this conversion may lead to precision loss due to the inherent imprecision of floating-point representations.

## Steps to Reproduce
1. Call `Decimal.parse_value(0.01)`.
2. Observe that the resulting `Decimal` does not accurately represent the value `0.01` instead gives`0.01000000000000000020816681711721685132943093776702880859375` which is inaccurate. 

## Expected Behavior
The function should convert the input to a string before creating the `Decimal` object (i.e., using `_Decimal(str(value))`). This approach ensures that decimal values are accurately converted, even if not provided as string, preserving their intended precision.

see [Strawberry implementation](https://github.com/strawberry-graphql/strawberry/blob/2af374359e9722980f5bc15b0d7158c8d54367b6/strawberry/schema/types/base_scalars.py#L25)

## Suggested Fix
Modify the `parse_value` function as follows:

```python
@staticmethod
def parse_value(value):
    try:
        return _Decimal(str(value))
    except Exception:
        return Undefined
```

mak626:
Fixes https://github.com/graphql-python/graphene/issues/1593

Example:-
`Decimal(0.01)` created a value `Decimal("0.01000000000000000020816681711721685132943093776702880859375")`

Additional:
The Decimal field now accepts string, int & float. Previously, it was just string & int

### DIFF
--- a/graphene/types/decimal.py
+++ b/graphene/types/decimal.py
@@ -1,7 +1,7 @@
 from decimal import Decimal as _Decimal
 
 from graphql import Undefined
-from graphql.language.ast import StringValueNode, IntValueNode
+from graphql.language.ast import StringValueNode, IntValueNode, FloatValueNode
 
 from .scalars import Scalar
 
@@ -22,13 +22,13 @@ class Decimal(Scalar):
 
     @classmethod
     def parse_literal(cls, node, _variables=None):
-        if isinstance(node, (StringValueNode, IntValueNode)):
+        if isinstance(node, (StringValueNode, IntValueNode, FloatValueNode)):
             return cls.parse_value(node.value)
         return Undefined
 
     @staticmethod
     def parse_value(value):
         try:
-            return _Decimal(value)
+            return _Decimal(str(value))
         except Exception:
             return Undefined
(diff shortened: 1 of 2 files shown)


graphene/types/decimal.py
```diff
@@ -22,13 +22,13 @@ def serialize(dec):
...
-        if isinstance(node, (StringValueNode, IntValueNode)):
+        if isinstance(node, (StringValueNode, IntValueNode, FloatValueNode)):
             return cls.parse_value(node.value)
         return Undefined
 
     @staticmethod
     def parse_value(value):
         try:
-            return _Decimal(value)
+            return _Decimal(str(value))
```
mak626, line 32: Using the same implementation as done at [Strawberry](https://github.com/strawberry-graphql/strawberry/blob/2af374359e9722980f5bc15b0d7158c8d54367b6/strawberry/schema/types/base_scalars.py#L25)

PR #1599, opened 2026-02-06
Support lazy type references for Union members

TomasBayer:
Fields, input fields, and arguments already support lazy type references via strings and lambdas to avoid circular imports. This PR adds the same support to member types of Unions. See this issue for details: https://github.com/graphql-python/graphene/issues/1600

Example:
```python
  class SomeUnion(Union):
      class Meta:
          types = ("package.Type", lambda: AnotherType) 
```

This change is non-breaking.


ahter (approved):
Thanks for the improvements @TomasBayer. This looks good to me.

